PR #2, opened 2022-03-10
Fix author and contributors

qdrn:
We usually publish package with Massa as author and developers as contributors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,8 +11,9 @@
     "test": "TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' mocha -r ts-node/register test/**/*.spec.ts",
     "test-client": "ts-node ./test/unit/client.ts"
   },
-  "author": "Evgeni Pirianov",
-  "license": "ISC",
+  "author": "Massa Labs <info@massa.net>",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -11,8 +11,9 @@
     "test": "TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' mocha -r ts-node/register test/**/*.spec.ts",
     "test-client": "ts-node ./test/unit/client.ts"
   },
-  "author": "Evgeni Pirianov",
-  "license": "ISC",
+  "author": "Massa Labs <info@massa.net>",
+  "contributors": ["Evgeni Pirianov", "Yvan Sraka <yvan@sraka.xyz>"],
+  "license": "MIT",
```
AureliaDolo, line 16: Here we choose that license, should we have the same for the massa repo ?

qdrn: No. For the Massa protocol we want to stay free to chose another license. We are still studying the possibility of constraining custodian staking. This may require a modification of standard licenses.

qdrn: Here it's a library to develop on Massa. We want to encourage people to build on it.

adrien-zinger: btw massa-sc-std, scripts and create are also MIT. I approve that change

PR #14, opened 2022-04-06
Use `npm` rather than `yarn`

yvan-sraka:
To please a GitHub Action that rely on `npm ci`


BatiGencho:
@yvan-sraka What is the rationale behind using npm as opposed to yarn ?

yvan-sraka:
> @yvan-sraka What is the rationale behind using npm as opposed to yarn ?

Fix the CI :)

BatiGencho:
Oh, yeah, ok. Agreed. I usually prefer yarn, but for that case, sure. Merging.

PR #16, opened 2022-04-11
Add some changes to make compile for react.

AurelienFT:


AurelienFT:
Current bug : 
```
Uncaught (in promise) TypeError: base58PublicKey.toBuffer is not a function
    at Function.<anonymous> (WalletClient.ts:223:1)
    at Generator.next (<anonymous>)
    at fulfilled (tslib.es6.js:73:1)
```
Apparently toBuffer doesn't exist on browser. Need to find a workaround

AurelienFT:
Potential fix : https://github.com/indutny/bn.js/issues/147

package.json
```diff
@@ -31,6 +31,7 @@
   "dependencies": {
     "@noble/secp256k1": "1.5.5",
     "axios": "^0.26.1",
+    "base58check": "^2.0.0",
     "bn.js": "^5.2.0",
     "bs58check": "^2.1.2",
```
qdrn, line 36: Is this one still used ? Maybe remove it from `package.json` otherwise.

qdrn: cc @AurelienFT

AurelienFT: My bad @BatiGencho cleaned it. ty

PR #26, opened 2022-06-28
Change operation encode publickey order.

AurelienFT:
With this PR : https://github.com/massalabs/massa/pull/2703 the publickey is now encoded always in first. I have made the change to match this.

This change will be up in the next testnet.


AurelienFT:
not uptodate.

PR #34, opened 2022-08-24
Feature/state

BatiGencho:
This PR is about extending the rpc filter for output events described here: https://github.com/massalabs/massa/issues/2857#issuecomment-1216812299


Eitu33 (approved):
LGTM but Damir asked for a bonus info in the return payload, not sure if it requires you to make a change in web3. Will keep you updated here once it is done, if you don't mind waiting before merging.

damip:
> LGTM but Damir asked for a bonus info in the return payload, not sure if it requires you to make a change in web3. Will keep you updated here once it is done, if you don't mind waiting before merging.

technically it doesn't require any changes because it's just extra info (retro-compatible). However it might allow implementing this in a simpler way

BatiGencho:
@damip Can I merge in that case the changes to master ?

damip:
> @damip Can I merge in that case the changes to master ?

yes.

Basically as an extra, we will soon also have a `"final": boolean` field inside the structures of the returned list of events, which will allow you to possibly simplify the internal implementation of this web3 feature, as a followup

PR #37, opened 2022-08-25
fixes and updates to the web3 lib

BatiGencho:
Fixes:

- proper async initialization of default and custom clients. Base account setter was async and not properly await-ed. Readme also changed
-  instead of the `candidate `key the getEvents rpc request data now uses the `is_final` key

Extensions:

- each IAddress has now also the createdInThread key which is needed for fast execution in some cases
- EventPoller has been modified to mimic proper websocket behaviour with regex filter and filter for events which have only happened after the last slot received. This way the event poller acts properly as a websocket delivering only new updates and filtering out old ones. This is irrelevant for a single events reading, but important where we have a contract calling itself repeatedly.

### DIFF
--- a/src/web3/EventPoller.ts
+++ b/src/web3/EventPoller.ts
@@ -10,6 +10,15 @@ import { Client } from "./Client";
 export const ON_MASSA_EVENT_DATA = "ON_MASSA_EVENT";
 export const ON_MASSA_EVENT_ERROR = "ON_MASSA_ERROR";
 
+const sortByThreadAndPeriod = (a: ISlot, b: ISlot): number => {
+	const periodOrder = a.period - b.period;
+	if (periodOrder === 0) {
+		const threadOrder = a.thread - b.thread;
+		return threadOrder;
+	}
+	return periodOrder;
+}
+
 /** Smart Contracts Event Poller */
 export class EventPoller extends EventEmitter {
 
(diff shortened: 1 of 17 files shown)


src/web3/EventPoller.ts
```diff
@@ -26,9 +29,47 @@ export class EventPoller extends EventEmitter {
...
+                if (this.lastSlot) {
+                    isAfterLastSlot = (event.context.slot.period > this.lastSlot.period) 
+					&& (event.context.slot.thread >= this.lastSlot.thread);
+                }
+				
+                return meetsRegex && isAfterLastSlot;
+            });
+
+			// sort after highest period and thread
+			const sortedByHighestThread = filteredEvents.sort((a, b) => {
```
qdrn, line 57: I believe you want to sort so that you obtain the latest slot. The latest slot is the one with the biggest period. If there are more than one slot with the biggest period, it is the slot with the biggest thread among those.
I believe you did the opposite (thread then period).

qdrn (changes requested):
LGTM with one comment.

src/web3/EventPoller.ts
BatiGencho: @qdrn Thanks for the comment. I have fixed that. Can you please do another review round ?

qdrn (approved):

qdrn:
You just need to update w.r.t. the main branch.

PR #52, opened 2022-11-16
50 - Add Args class to serialize arguments into bytes (u32, string)

Thykof:


src/utils/arguments.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    let value = BigInt(0);
+
+    // encoding is little endian
+    for (let i = 3; i >= 1; --i) {
+      value = (value | BigInt(byteArray[this.offset + i])) << BigInt(8);
+    }
+
+    value = value | BigInt(byteArray[this.offset]);
+    this.offset += 4;
```
gregLibert, line 64: You should remove the magic value here. My proposal :

```suggestion
   numBytes = sizeof<u32>();
    if (byteArray.length - this.offset < numBytes) {
      return BigInt(NaN);
    }

    let value = BigInt(0);

    // encoding is little endian
    for (let i = NumBytes-1; i > 0; --i) {
      value = (value | BigInt(byteArray[this.offset + i])) << BigInt(8);
    }

    value = value | BigInt(byteArray[this.offset]);
    this.offset += numBytes;
```

src/utils/arguments.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  // Setter
+
+  /**
+   *
+   * @param {BigInt} bigInt
+   * @return {Args}
+   */
+  addU32(bigInt): Args {
+    const u32 = BigInt.asUintN(32, bigInt);
```
gregLibert, line 77:
```suggestion
    const u32 = BigInt.asUintN(/* num of significant bits */ 32, bigInt);
```

src/utils/arguments.ts
```diff
@@ -0,0 +1,144 @@
...
+    const byteArray = new Uint8Array(4);
+    for (let i = 0; i < byteArray.length; i++) {
+      // extracts the value of the first, second, third or forth byte by moving
+      // the _select all bits_ (0xFF) mask
+      const targetByte = u32 & (BigInt(0xff) << BigInt(8 * i));
+      // reduces the extracted value to one byte by removing
+      // already encoded and zeroed byte
+      const byte = targetByte >> BigInt(8 * i);
+      byteArray[i] = Number(byte);
+    }
```
gregLibert, line 87:
```suggestion
    const byteArray = new Uint8Array(4);
    const byteMask = BigInt(0xff); // extracts all the bits of a given byte
    for (let i = 0; i < byteArray.length; i++) {
      // extracts the value of the first, second, third or forth byte by moving the mask
      const targetByte = u32 & (byteMask << BigInt(8 * i));
      // reduces the extracted value to targeted byte by removing lower masked bytes
      const byte = targetByte >> BigInt(8 * i);
      byteArray[i] = Number(byte);
    }
```

src/utils/arguments.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  // Setter
+
+  /**
+   *
+   * @param {BigInt} bigInt
+   * @return {Args}
+   */
+  addU32(bigInt): Args {
```
gregLibert, line 76: No argument type ?

src/utils/arguments.ts
```diff
@@ -0,0 +1,144 @@
...
+  /**
+   * Adds an argument to the serialized byte string if the argument is an
+   * instance of a handled type (String of 4294967295 characters maximum)
+   *
+   * @param {string} arg the argument to add
+   *
+   * @return {Args} the modified Arg instance
+   */
+  addString(arg: string): Args {
+    if (typeof arg != "string") {
```
gregLibert, line 103: I guess if this function argument type is string, the typeof is also string no ?

src/utils/arguments.ts
```diff
@@ -0,0 +1,144 @@
...
+      console.warn(
+        "Wrong arg type, required a string with max length 4294967295"
+      );
+    } else {
+        this.addU32(BigInt(arg.length));
```
gregLibert, line 108: Why not checking the size of the string ?

src/utils/arguments.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  // Utils
+
+  /**
+   * Returns a byte string.
+   *
+   * @param {Uint8Array} byteString
+   * @return {string}
+   */
+  toByteString(byteString: Uint8Array): string {
```
gregLibert, line 123: Why not static ? Private ?

gregLibert, line 123: The argument is not a byte string :-)
```suggestion
  toByteString(bytes: Uint8Array): string {
```

gregLibert (changes requested):
:no_good: The lisibility of this code can be easily improved by:
- removing magic numbers
- clarifying variable names
- adding new variables
- changing function to static/private

src/utils/arguments.ts
Thykof: - when I use BigInt it says me Argument of type 'BigInt' is not assignable to parameter of type 'bigint'.ts(2345) at the line bellow
- when I use bigint it says: Cannot find name 'Bigint'. Did you mean 'BigInt'?ts(2552)


I can do 
```
addU32(bigInt: number): Args {
    const u32 = BigInt.asUintN(32, Number(bigInt));
```

src/utils/arguments.ts
Thykof: you're right!

src/utils/arguments.ts
Thykof: there is no sizeof in typescript

src/utils/arguments.ts
Thykof: ok let's do this and slice the characters that are too much

src/utils/arguments.ts
Thykof: let's mark it as private

src/utils/arguments.ts
gregLibert: :grin:

src/utils/arguments.ts
gregLibert: If you don't use this, you should make it private :-)

gregLibert (approved):
:clap: Very first step.

PR #83, opened 2022-12-22
added ws support

BatiGencho:
WS implementation with a cliient, default configs, reconnect mechanisms, pinging and events. Message parsing is also included. Tests added.


src/web3/WsBlockSubClient.ts
```diff
@@ -0,0 +1,75 @@
...
+		const messageStr = bin2String(data);
+		let parsedMsg: Object = null;
+		try {
+			parsedMsg = JSON.parse(messageStr);
+			if (parsedMsg["result"]) {
+				this.subId = parseInt(parsedMsg["result"]);
+			}
+			if (parsedMsg["params"] && parsedMsg["params"]["result"] && parsedMsg["params"]["subscription"] === this.subId) {
+				if (this.onNewBlockHandler) {
+					this.onNewBlockHandler(parsedMsg["params"]["result"] as ISubscribeNewBlocksMessage);
```
AurelienFT, line 68: Block and FilledBlock are differents. FilledBlock have the operations field but Block only have operation_ids

AurelienFT (commented):
Why did you made a new client and not group them ? This will cause a lot of compatibility error.

src/web3/WsBlockSubClient.ts
BatiGencho: I thought it might be easier for the user to work with separate targeted clients, but you might be right in saying that one client would be better. I will change that.

BatiGencho: I have updated now all of that.

AurelienFT (approved):
Very good job. Thank you !

PR #86, opened 2023-01-05
updated json rpcs for testnet 18

BatiGencho:


src/interfaces/JsonRpcMethods.ts
```diff
@@ -3,7 +3,7 @@ export enum JSON_RPC_REQUEST_METHOD {
 	GET_STATUS = "get_status",
 	GET_ADDRESSES = "get_addresses",
 	SEND_OPERATIONS = "send_operations",
-	GET_BLOCKS = "get_block",
+	GET_BLOCKS = "get_blocks",
```
AurelienFT, line 6: Be carefule now this endpoints takes a vector instead of one block id.

BatiGencho: fixed, thank you!

AurelienFT (approved):

PR #87, opened 2023-01-09
added read value, improvements in sc methods

BatiGencho:
Read values were added to readScOperation, some smart contract methods reworked and thus made easier for users, Doc enhanced, test example fixed


src/interfaces/IExecuteReadOnlyData.ts
```diff
@@ -0,0 +1,14 @@
...
+import { IReadOperationResult } from "./IReadOperationResult";
+import { ISlot } from "./ISlot";
+
+export interface IExecuteReadOnlyData {
+    /// The slot at which the read-only execution occurred.
+	executed_at: ISlot;
+    /// The result fo the read-only execution.
+	result: IReadOperationResult;
+    /// The output events generated by the read-only execution.
+	output_events: Map<string, IEvent>;
```
AurelienFT, line 11: This should be like in `IContractReadOperationData` an `Array<IEvent>` no ?

src/interfaces/IExecuteReadOnlyResponse.ts
```diff
@@ -1,13 +1,6 @@
...
 export interface IExecuteReadOnlyResponse {
-    /// The slot at which the read-only execution occurred.
-	executed_at: ISlot;
-    /// The result fo the read-only execution.
-	result: string | null;
-    /// The output events generated by the read-only execution.
-	output_events: Map<string, IEvent>;
-    /// The gas cost for the execution
-    gas_cost: number;
+	returnValue: Uint8Array;
```
AurelienFT, line 4: Why does this not match the API and have a wrapper structure like that ?

AurelienFT (commented):
Nice refactor, good improvements

src/interfaces/IExecuteReadOnlyResponse.ts
AurelienFT: Answered on DIscord : The api response is in the info: IExecuteReadOnlyData. The idea is to have the returnValue ready for the user to use rather than to extract it from the IExecuteReadOnlyData. If one needs some further data contained inside IExecuteReadOnlyData, the info can be always inspected.

AurelienFT (approved):

PR #100, opened 2023-02-15
Feature/docs improved

BatiGencho:
- added linter (eslint)
- removed tslint
- added docs
- formattings
- pipelines (ci actions)
- removed old files/tests

PR #101, opened 2023-02-15
Feature/docs improved fix prettier

Thykof:
I think you forgot to add prettier config massalabs in package.json


Thykof:
this PR has been integrated in https://github.com/massalabs/massa-web3/pull/100

PR #109, opened 2023-02-21
added units and utils

BatiGencho:


BatiGencho:
@gregLibert I agree, do you want me to do that ?

gregLibert:
> @gregLibert I agree, do you want me to do that ?

Let's the inno team react to this, because if we do it, we will need to do it on Thyra side as well..

Thykof:
I think it's a good idea.

peterjah:
Personnaly i don't really like these wrapping, i would say its much easier to use directly bignumbers to handle all kind of amounts (fees ,coin etc...),
Then we can have constants as @gregLibert mentionned like milliMassa or event satoMassa :)

peterjah:
we should take inspiration from what does ethers.js
https://docs.ethers.org/v4/api-utils.html#ether-strings-and-wei

peterjah:
in addition, i tink we should prevent user to use number type to represent amounts. Here we do the opposite by providing "toNumber" method

BatiGencho:
@qdrn @peterjah @Thykof @gregLibert Do we all agree that we want to do something similar to ethers.js ? And we all say we follow the metric system: milli, nano ,micro ?

gregLibert:
> @qdrn @peterjah @Thykof @gregLibert Do we all agree that we want to do something similar to ethers.js ? 

I guess we should clarify what is "something similar to ether.js".

I shared code sample of what I have in mind:
- a specific type for amount express in massa
- constant usage to make the code easier to understand
- explicite function to convert the amount to string

Obvisouly things are missing, but when I see things such as:

```js
console.log(utils.formatEther(wei, {commify: true, pad: true}));
// "1,000.000000000000000000" 
```
i'm not necessarly ont the same page as ether.js because this way of writing number is the american one, not necessarly the way the enduser want it.

> And we all say we follow the metric system: milli, nano ,micro ?

The crypto eco system is cool, but when you have international standards, I guess you should use them...

leoloco:
In my opinion, everything should be expressed as the smallest denomination of the Massa token, the equivalent of "wei" for ethereum. Then we, or the community can develop and use as many wrappers/converters as they wish, depending on their needs. But it should be clearly defined and documented that any amount of Massa, rather it bee fees gas roll or amount is expressed in the smallest denomination.

Thykof (changes requested):

src/web3/MassaAmount.ts
```diff
@@ -0,0 +1,65 @@
...
+const MASSA_SCALING_FACTOR = 9;
+
+export class MassaAmount {
+  private value: BigNumber;
+  private unit: MASSA_UNIT;
+
+  constructor(value: BigNumber | number | string, unit: MASSA_UNIT) {
+    // truncate value to smallest massa unit
+    this.value = new BigNumber(new BigNumber(value).toFixed(MASSA_SCALING_FACTOR));
+    this.unit = unit;
```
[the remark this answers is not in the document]
Thykof: here is what I propose https://github.com/massalabs/massa-web3/pull/119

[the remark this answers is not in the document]
BatiGencho: accepted

examples/smartContracts/deployer.ts
```diff
@@ -118,15 +118,15 @@ export const deploySmartContracts = async (
...
+      if (contract.coins.getValue().isGreaterThan(0)) {
         datastore.set(
           new Uint8Array(
             new Args()
               .addU64(BigInt(i + 1))
               .addUint8Array(u8toByte(1))
               .serialize(),
           ),
-          u64ToBytes(BigInt(contract.coins.toNumber())), // scaled value to be provided here
+          u64ToBytes(BigInt(contract.coins.toCoins().getValue().toNumber())), // scaled value to be provided here
```
peterjah, line 129: so what happen if the amount doesnt fit on number ?. Here you are coverting a bignumber to number an then to bigInt.. You should use string instead of number.

src/web3/MassaAmount.ts
```diff
@@ -0,0 +1,78 @@
...
+    return this.unit;
+  }
+
+  public toRolls(): MassaAmount {
+    const valueInMassa = this.toUnit(MASSA_UNIT.MASSA);
+    return new MassaAmount(valueInMassa.value.toFixed(0), MASSA_UNIT.MASSA);
+  }
+
+  public toGas(): MassaAmount {
+    return new MassaAmount(this.value.toFixed(0), this.unit);
```
peterjah, line 50: `return this` should work here ?

src/web3/MassaAmount.ts
```diff
@@ -0,0 +1,78 @@
...
+    return this.value;
+  }
+
+  public getUnit(): MASSA_UNIT {
+    return this.unit;
+  }
+
+  public toRolls(): MassaAmount {
+    const valueInMassa = this.toUnit(MASSA_UNIT.MASSA);
+    return new MassaAmount(valueInMassa.value.toFixed(0), MASSA_UNIT.MASSA);
```
peterjah, line 46: `return this.toUnit(MASSA_UNIT.MASSA);` it is weird to initialize a new massaAmount object here..

test/conversion.spec.ts
```diff
@@ -0,0 +1,98 @@
+import 'mocha';
```
peterjah, line 1: Thank you for adding tests 😄

peterjah (changes requested):

peterjah:
> In my opinion, everything should be expressed as the smallest denomination of the Massa token, the equivalent of "wei" for ethereum. Then we, or the community can develop and use as many wrappers/converters as we wish, depending on the needs. But it should be clearly defined and documented that any amount of Massa, rather it be fees roll or amount is expressed in the smallest denomination.

+1

src/web3/MassaAmount.ts
Thykof: no because of toFixed(0à

test/conversion.spec.ts
BatiGencho: no worries, my pleasure :)

src/web3/MassaAmount.ts
BatiGencho: @peterjah Same argumentation as above: this is needed because of .toFixed(0)

src/web3/MassaAmount.ts
BatiGencho: Exactly

examples/smartContracts/deployer.ts
BatiGencho: I fixed this part as you suggested @peterjah

peterjah (approved):

src/web3/MassaAmount.ts
```diff
@@ -0,0 +1,78 @@
+import BigNumber from 'bignumber.js';
+
+export enum MASSA_UNIT {
```
gregLibert, line 3: [Nit-picking]As type script is a typed language, do we need to write anum value in capital letter? I have the feeling that the code is yelling at me :-)

test/conversion.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+import { expect } from 'chai';
+import { MassaAmount, MASSA_UNIT } from '../src';
+
+describe('MassaAmount Class', () => {
+  it('classical unit test conversion cases', () => {
+    const amount = new MassaAmount(1, MASSA_UNIT.MASSA);
```
gregLibert, line 7: I would have created a Massa(), MilliMassa() and NanoMassa() and add basic operators to MassaAmount to offer a way to write cleaner code such as:

```typescript
const amount = 1 * Massa()
```
Furthermore, calculus on amount would have been possible.

test/conversion.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+    expect(amount.toString()).to.equal('1 MASSA');
+
+    const milliMassa = amount.toUnit(MASSA_UNIT.MILLI_MASSA);
+    expect(milliMassa.toString()).to.equal('1000 MILLI_MASSA');
```
gregLibert, line 11: I would have define a custom toString function to have more friendly currency name:
- MILLI_MASSA => mMASSA
- MICRO_MASSA => µMASSA
- NANO_MASSA => nMASSA

gregLibert (commented):
Some suggestions that can be addressed here or in another issue.

Also, there are many CI verification errors that should be addressed at some point (here or in another problem).

peterjah:
after thinking together, here are the specs:

- represent all amounts with the smallest unit: nano massa
- use BigNumber without wrapping it in a class
- no need for a toString function, BigNumber has already one, because no need for human representation with the unit
- create function helper to convert BigNumber (nano massa) to massa coin (human readable string)
- no need for milli massa nor micro massa.

BatiGencho:
> 
Comments on my side: 

@gregLibert I agree with removing micro and milli. However, my question is : 
How is the user expected to input his coins e.g. new BigNumber(12300000000)?

gregLibert:
> @gregLibert I agree with removing micro and milli. However, my question is :
> How is the user expected to input his coins e.g. new BigNumber(12300000000)?

I see two possibilities:

```javascript

let amount = 12_300_000_000n;
amount = 12_300n * mMassa();
```

BatiGencho:
> 

I dont understand. What is the n in 12_300_000_000n meant to symbolize ? This is not a js number, is it ?

Thykof:
the n creates a bigint 
image

BatiGencho:
@Thykof @peterjah @gregLibert  I wanted to hear what your thoughts are on doing this:

I extended the BigNumber prototype with extra functionalities and now we can say, for example:

    `const amount = (new BigNumber(1.5)).asMassa();`
    
    which will under the hood convert the 1.5 into nanoMassa for direct use by us.
    
    We can also have :
    
       `const amount = (new BigNumber(15000000000)).asNanoMassa(); `
       
       which will not preserve the number as nano massa. 
       
       Would you be all happy with such BigNumber prototype extensions ?

peterjah:
> `const amount = (new BigNumber(1.5)).asMassa();`

yes and i think it should return a string. so no overflow issues and because it will mostly be used for logging purpose
> 
> We can also have :
> 
> `const amount = (new BigNumber(15000000000)).asNanoMassa(); `
> 
this one should be useless because new BigNumber("15000000000") is already nanomassa

gregLibert:
> this one should be useless because new BigNumber("15000000000") is already nanomassa

Useless, so not necessarly to include in code sample, but at least the code reader understand it ;-).

> @Thykof @peterjah @gregLibert I wanted to hear what your thoughts are on doing this:

It's cooler, thanks. I know I was the one to talk about using a function, but do you think that using a constant won't be cleaner ?


``` typescript
...

const amount = 1n * µMassa; // µMassa = 1000n
...
```

gregLibert:
> Carefull to not mixing types, BigInt is not he same as BigNumbers

Good point! Why using bigNumber if in fact our format is a u64 ?

examples/smartContracts/index.ts
```diff
@@ -227,14 +227,16 @@ const pollAsyncEvents = async (
     const callOperationId = await web3Client
       .smartContracts()
       .callSmartContract({
-        fee: 0,
-        maxGas: 10_500_000,
-        coins: new MassaCoin(0),
+        fee: fromMAS(0),
```
peterjah, line 230: `0n` is maybe more direct?

BatiGencho: yep, thx. changed

examples/smartContracts/index.ts
```diff
@@ -140,7 +140,7 @@ const pollAsyncEvents = async (
     console.log(
       `Deployer Wallet Address: ${
         deployerAccount.address
-      } with balance (candidate, final) = (${deployerAccountBalance?.candidate.rawValue()}, ${deployerAccountBalance?.final.rawValue()})`,
+      } with balance (candidate, final) = (${deployerAccountBalance?.candidate.toString()}, ${deployerAccountBalance?.final.toString()})`,
```
peterjah, line 143: maybe it will be nice to use the `toMas()` here ?!

peterjah (approved):

examples/smartContracts/index.ts
BatiGencho: sure, done, added

Thykof:
we should update https://github.com/massalabs/massa-standards/issues/20

README.md
```diff
@@ -765,6 +765,47 @@ const data: IExecuteReadOnlyResponse = await web3Client
...
+
+All Massa values that are being used or returned by web3 (gas, fees, coins and rolls) are expressed via BigInt's. Massa-web3 has however a few convenience methods and converters that might come handy. Below is a summary and some examples of the latter:
+
+- **Rolls**: expressed in BigInt's. For Rolls there is no metric system as rolls are unit-less. 10 rolls is to be represented by a BigInt containing 10. Example:
+```ts
+const rolls = BigInt(10);
+// or. ...
+const rolls = 10n;
+```
+- **Gas/MaxGas**: expressed in BigInt's. For Gas/MaxGas there is no metric system as rolls are unit-less. The gas represents the computational units required for a given operation to be executed by the network. Example:
```
Thykof, line 778: does gas also unit-less?

```suggestion
- **Gas/MaxGas**: expressed in BigInt's. For Gas/MaxGas there is no metric system as gas is unit-less. The gas represents the computational units required for a given operation to be executed by the network. Example:
```

src/utils/converters.ts
```diff
@@ -2,20 +2,40 @@ import BigNumber from 'bignumber.js';
...
-export const fromMAS = (inputAmount: number | string): BigNumber => {
-  const amount = new BigNumber(inputAmount);
+/**
+ * Convert any value expressed in nano Massa back to Massa
+ * @param amountInNanoMassa any amount in nanoMassa
+ * @returns BigNumber representing the amount expressed in Massa
+ */
+export const toMAS = (
+  amountInNanoMassa: number | string | BigNumber | bigint,
+): BigNumber => {
```
Thykof, line 28: why toMAS returns BigNumber whereas fromMAS returns bigint?

examples/smartContracts/index.ts
```diff
@@ -208,8 +210,8 @@ const pollAsyncEvents = async (
     ).start();
     const args = new Args();
     const result = await web3Client.smartContracts().readSmartContract({
-      fee: 0,
-      maxGas: 700000,
+      fee: 0n,
+      maxGas: BigInt(700000),
```
Thykof, line 214: why not 700_000n? same line 233
```suggestion
      maxGas: 700_000n,
```

examples/wallet/index.ts
```diff
@@ -107,8 +108,8 @@ if (!receiverPrivateKey) {
 
     // send from base account to receiver
     const txId = await web3Client.wallet().sendTransaction({
-      amount: '1',
-      fee: 0,
+      amount: fromMAS(1),
+      fee: fromMAS(0),
```
Thykof, line 112: why not 0n or 0? same line 159?
```suggestion
      fee: 0n,
```

src/web3/SmartContractsClient.ts
```diff
@@ -184,7 +184,7 @@ export class SmartContractsClient
   ): Promise<IContractReadOperationResponse> {
     // request data
     const data = {
-      max_gas: readData.maxGas,
+      max_gas: Number(readData.maxGas),
```
Thykof, line 187: the readme is saying that max gas is expressed in BigInt, why here you concert to Number?

Thykof (changes requested):

src/web3/SmartContractsClient.ts
BatiGencho: because that data is the payload I am sending to the node over json-rpc. it needs to be a number (u64 in rust), otherwise the call fails.

examples/wallet/index.ts
BatiGencho: sure, I have changed it. Thakks

examples/smartContracts/index.ts
BatiGencho: changed, thanks.

src/utils/converters.ts
BatiGencho: because e.g. 300_000_000  is 0.3 massa. and 0.3 can go into a BigNumber.

BatiGencho:
> we should update [massalabs/massa-standards#20](https://github.com/massalabs/massa-standards/issues/20)

https://github.com/massalabs/massa-standards/pull/28 I have attached the MR to that issue.

Thykof (approved):

PR #111, opened 2023-02-21
improved polling on the deployer example, datastore entries etc.

BatiGencho:


examples/smartContracts/index.ts
```diff
@@ -8,17 +8,106 @@ import {
...
 const path = require('path');
 const chalk = require('chalk');
 const ora = require('ora');
 
+const MASSA_EXEC_ERROR = 'massa_execution_error';
+
 const DEPLOYER_SECRET_KEY =
   'S1NA786im4CFL5cHSmsGkGZFEPxqvgaRP8HXyThQSsVnWj4tR7d';
 
+export async function withTimeoutRejection<T>(
```
Thykof, line 31: it looks like duplicated code with the sc-deployer. maybe you should refactor code to avoid duplicating functions

Thykof (changes requested):

examples/smartContracts/index.ts
BatiGencho: I will export that from web3 under utils and make it available for the sc-depliyor to import it. Sounds ok ?

Thykof: yes

Thykof (approved):
maybe the example should be in massa-sc-examples?

peterjah (approved):

PR #116, opened 2023-02-24
fix axios version > 1.0

BatiGencho:


BatiGencho:
[https://github.com/massalabs/massa-eth-bridge/issues/7](url)

package.json
```diff
@@ -60,7 +60,7 @@
         "@noble/ed25519": "^1.7.3",
         "@noble/hashes": "^1.2.0",
         "@types/ws": "^8.5.4",
-        "axios": "^1.3.3",
+        "axios": "^0.26.1",
```
Thykof, line 63: you should create an issue to install again the latest version of axios

Thykof (approved):

package.json
Thykof: I did see https://github.com/axios/axios/issues/5262 ignore my comment

PR #126, opened 2023-04-05
added typed arrays ser & deser

BatiGencho:
Added typed arrays serialization and deserialization.
Original issue for dusa: [issue](https://github.com/massalabs/massa-web3/issues/127)

### DIFF
--- a/src/utils/arguments.ts
+++ b/src/utils/arguments.ts
@@ -20,6 +20,7 @@ import {
   deserializeObj,
   bytesToSerializableObjectArray,
   nativeTypeArrayToBytes,
+  bytesToNativeTypeArray,
 } from './serializers';
 
 /**
@@ -30,8 +31,8 @@ import {
  * @remark In Assemblyscript the latter are all native types
  */
 export enum TypedArrayUnit {
+  BOOL,
   U8,
-  U16,
   U32,
   U64,
   I32,
@@ -124,6 +125,13 @@ export class Args {
     const value = bytesToU64(this.serialized, this.offset);
     this.offset += 8;
     return value;
+  }
+
+  /**
+   * Returns the deserialized boolean
+   */
+  nextBool(): boolean {
+    return !!this.serialized[this.offset++];
   }
 
   /**
@@ -220,6 +228,27 @@ export class Args {
     return value;
   }
 
+  /**
+   * @returns the next array of object that are native type
+   */
+  nextNativeTypeArray<T extends TypedArrayUnit>(typedArrayType: T): T[] {
+    const length = this.nextU32();
+    if (this.offset + length > this.serialized.length) {
+      throw new Error("can't deserialize length of array from given argument");
+    }
+
+    const bufferSize = length;
+
+    if (bufferSize === 0) {
+      return [];
+    }
+
+    const buffer = this.getNextData(bufferSize);
+    const value = bytesToNativeTypeArray<T>(buffer, typedArrayType);
+    this.offset += bufferSize;
+    return value;
+  }
+
   // Setter
 
   /**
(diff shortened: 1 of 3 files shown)


src/utils/arguments.ts
```diff
@@ -351,6 +415,25 @@ export class Args {
...
+   * @remarks
+   * If the type of the values of the array is not native type, this will serialize the pointers, which is certainly not
+   * what you want. You can only serialize properly array of native types or array of `Serializable` object.
+   *
+   * @see {@link addSerializableObjectArray}
+   *
+   * @param arg - the argument to add
+   * @returns the modified Arg instance
+   */
+  addNativeTypeArray<T extends TypedArrayUnit>(arg: any[], type: T): Args {
```
peterjah, line 430: Imho the function signature shoud be:
`addNativeTypeArray<T>(arg: T[]): Args {`
Then you should be able to handle different types like this 
```
  switch (idof<T>()) {
    case idof<i8[]>(): // ArrayI8;
    case idof<u8[]>(): // ArrayU8;
    case idof<i16[]>(): // ArrayI16;
    case idof<u16[]>(): // .ArrayU16;
    ```
etc...
 It have the benefit to infer the intern type of the array instead of explicitely setting it everytime. and thus make the code cleaner.
 Not tested of course.. so its only a improvement suggestions!

**Comment by peterjah** [comment]
Also a reminder: we have to implement the equivalent in TS in massa web3!

File: src/utils/serializers/arrays.ts [hunk 2]
```diff
@@ -39,3 +69,174 @@ export function deserializeObj<T extends ISerializable<T>>(
...
+        result.push(dataView.getBigInt64(byteOffset, true) as unknown as T);
+        byteOffset += size;
+        break;
+      case TypedArrayUnit.U32:
+        result.push(dataView.getUint32(byteOffset, true) as unknown as T);
+        byteOffset += size;
+        break;
+      case TypedArrayUnit.U64:
+        result.push(dataView.getBigUint64(byteOffset, true) as unknown as T);
+        byteOffset += size;
```
Thykof, line 237: this instruction is executed in every cases of the switch, consider doing it one time just after the switch

Thykof (changes requested):

BatiGencho:
> 

What do you mean by this ? This is in massa-web3

src/utils/serializers/arrays.ts
```diff
@@ -1,7 +1,37 @@
+/* eslint-disable no-case-declarations */
+
+const MAX_STRING_CHARS = 100;
```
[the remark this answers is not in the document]
BatiGencho: I have limited the number of chars of a string for an array of strings we could support not to cause v8 out of memory. The problem with arrays as you pointed out below is that they are of unknown size so 2 options were available: 
- to use the max number which is hardcoded btw. , but if the array has several elements this would cause v8 out of memory exception
- to put an upper bound as I did (we could increase it for sure), but keeping it below the max 4789655 (the max value).

src/utils/serializers/arrays.ts
```diff
@@ -1,7 +1,37 @@
...
+
+// Lookup table for native types
+const getDatatypeSize = (typedArrayType: TypedArrayUnit): number => {
+  switch (typedArrayType) {
+    case TypedArrayUnit.STRING:
+      return MAX_STRING_CHARS;
```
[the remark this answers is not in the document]
BatiGencho: right, see my comment above reg. this point.

src/utils/serializers/arrays.ts
BatiGencho: > this instruction is executed in every cases of the switch, consider doing it one time just after the switch

Thanks, fixed that. Good point!

Thykof (approved):

peterjah:
> Also a reminder: we have to implement the equivalent in TS in massa web3!

lol sorry i though it a PR on AS 😄

peterjah:
> > 
> 
> What do you mean by this ? This is in massa-web3

Yeah sorry about his. I think i will stay away from PR reviews about AS and smart contract tooling from now... 🥲 I am working on totally different topics.

peterjah (approved):

Issue #152, opened 2023-05-02
Check and update if necessary the web3 documentation


Elli610:
- [x] https://github.com/massalabs/massa-web3/pull/161
- [x] https://github.com/massalabs/massa-web3/pull/160
- [x] https://github.com/massalabs/massa-web3/pull/190
- [x] https://github.com/massalabs/massa-web3/pull/180
- [x] https://github.com/massalabs/massa-web3/pull/179
- [x] https://github.com/massalabs/massa-web3/pull/159
- [x] https://github.com/massalabs/massa-web3/pull/157
- [x] https://github.com/massalabs/massa-web3/pull/155
- [x] https://github.com/massalabs/massa-web3/pull/153

PR #160, opened 2023-05-02
Update SmartContractsClient.ts

Elli610:
related to #152

### DIFF
--- a/src/web3/SmartContractsClient.ts
+++ b/src/web3/SmartContractsClient.ts
@@ -1,3 +1,11 @@
+/**
+ * This file includes the implementation for the {@link SmartContractsClient} class. This class provides methods for interacting with smart contracts
+ * in the Massa blockchain. Such methods include {@link SmartContractsClient#deploySmartContract|deploying}, {@link SmartContractsClient#callSmartContract|calling},
+ * and {@link SmartContractsClient#readSmartContract|reading} smart contracts, as well as retrieving smart contract {@link SmartContractsClient#getFilteredScOutputEvents|events}
+ * and {@link SmartContractsClient#getContractBalance|balances}.
+ *
+ * @module SmartContractsClient
+ */
 import { EOperationStatus } from '../interfaces/EOperationStatus';
 import { IAccount } from '../interfaces/IAccount';
 import { IAddressInfo } from '../interfaces/IAddressInfo';
@@ -30,14 +38,14 @@ const TX_POLL_INTERVAL_MS = 10000;
 const TX_STATUS_CHECK_RETRY_COUNT = 100;
 
 /**
- * Smart Contracts Client object enables smart contract deployment, calls and streaming of events
+ * Smart Contracts Client object enables smart contract deployment, calls and streaming of events.
  */
 export class SmartContractsClient
   extends BaseClient
   implements ISmartContractsClient
 {
   /**
-   * Constructor for {@link SmartContractsClient} objects
+   * Constructor for {@link SmartContractsClient} objects.
    */
   public constructor(
     clientConfig: IClientConfig,
@@ -61,15 +69,16 @@ export class SmartContractsClient
 
   /**
    * Deploy a smart contract on th massa blockchain by creating and sending
-   * an operation containing byte code.
+   * an operation containing byte code
    *
    * @remarks
-   * If no executor is provided, the default wallet account will be used.
+   * If no executor is provided, the default wallet account from the provided {@link WalletClient}
+   * will be used.
    *
-   * @param contractData - A IContractData object containing the deployment data
-   * @param executor - An optional IAccount object containing the account to use for the deployment
+   * @param contractData - The deployment contract data.
+   * @param executor - The account to use for the deployment.
    *
-   * @returns A promise that resolves to a string containing the operation id of the deployment operation
+   * @returns A promise that resolves to the operation ID of the deployment operation.
    */
   public async deploySmartContract(
     contractData: IContractData,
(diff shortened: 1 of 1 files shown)


src/web3/SmartContractsClient.ts
```diff
@@ -121,7 +137,16 @@ export class SmartContractsClient
...
-  /** call smart contract method */
+  /**
+   * Calls a smart contract method
+   *
+   * @remarks
+   * If no executor is provided, the default wallet account will be used.
+   *
+   * @param callData - A ICallData object containing the call data
+   * @param executor - An optional IAccount object containing the account to use for the call
+   * @return A promise that resolves to a string containing the operation id of the call operation
```
0xSwapFeeder, line 148:
```suggestion
   *
   * @return A promise that resolves to a string containing the operation id of the call operation
```

src/web3/SmartContractsClient.ts
```diff
@@ -179,7 +204,13 @@ export class SmartContractsClient
     return opIds[0];
   }
 
-  /** read smart contract method */
+  /**
+   * Read a smart contract method
+   *
+   * @param readData - A IReadData object containing useful data for the operation
+   * @return A promise that resolves to a IContractReadOperationResponse object
```
0xSwapFeeder, line 211:
```suggestion
   *
   * @return A promise that resolves to a IContractReadOperationResponse object
```

src/web3/SmartContractsClient.ts
```diff
@@ -229,7 +260,12 @@ export class SmartContractsClient
     } as IContractReadOperationResponse;
   }
 
-  /** Returns the balance of the smart contract  */
+  /**
+   * Returns the balance of the smart contract
+   *
+   * @param address - The address of the smart contract
+   * @return A promise that resolves to a IAddressInfo object containing the balance of the smart contract
```
0xSwapFeeder, line 267:
```suggestion
   *
   * @return A promise that resolves to a IAddressInfo object containing the balance of the smart contract
```

src/web3/SmartContractsClient.ts
```diff
@@ -241,7 +277,12 @@ export class SmartContractsClient
     } as IBalance;
   }
 
-  /** get filtered smart contract events */
+  /**
+   * Get filtered smart contract events
+   *
+   * @param eventFilterData - A IEventFilter object containing the filter
+   * @return A promise that resolves to an array of IEvent objects containing the filtered events
```
0xSwapFeeder, line 284:
```suggestion
   *
   * @return A promise that resolves to an array of IEvent objects containing the filtered events
```

src/web3/SmartContractsClient.ts
```diff
@@ -271,7 +312,13 @@ export class SmartContractsClient
     }
   }
 
-  /** Read-only smart contracts */
+  /**
+   * Execute a read-only smart contracts
+   *
+   * @param contractData - A IContractData object containing the contract data
+   * @return A promise which resolves to a IExecuteReadOnlyResponse object containing the result
```
0xSwapFeeder, line 319:
```suggestion
   *
   * @return A promise which resolves to a IExecuteReadOnlyResponse object containing the result
```

src/web3/SmartContractsClient.ts
```diff
@@ -319,6 +366,12 @@ export class SmartContractsClient
     } as IExecuteReadOnlyResponse;
   }
 
+  /**
+   * Get the status of a specific operation
+   *
+   * @param opId - The operation id
+   * @return A promise that resolves to a EOperationStatus enum value
```
0xSwapFeeder, line 373:
```suggestion
   *
   * @return A promise that resolves to a EOperationStatus enum value
```

src/web3/SmartContractsClient.ts
```diff
@@ -338,6 +391,13 @@ export class SmartContractsClient
     return EOperationStatus.INCONSISTENT;
   }
 
+  /**
+   * Get the status of a specific operation and wait until it reaches the required status
+   *
+   * @param opId - The required operation id
+   * @param requiredStatus - The required status
+   * @return A promise that resolves to a EOperationStatus enum value
```
0xSwapFeeder, line 399:
```suggestion
   *
   * @return A promise that resolves to a EOperationStatus enum value
```

0xSwapFeeder (changes requested):

src/web3/SmartContractsClient.ts
```diff
@@ -29,11 +29,16 @@ const MAX_READ_BLOCK_GAS = BigInt(4_294_967_295);
 const TX_POLL_INTERVAL_MS = 10000;
 const TX_STATUS_CHECK_RETRY_COUNT = 100;
 
-/** Smart Contracts Client which enables compilation, deployment and streaming of events */
+/**
+ * Smart Contracts Client object enables compilation, deployment and streaming of events
```
0xSwapFeeder, line 33: Compilation ? And what about smart contract calls ?

0xSwapFeeder (changes requested):

src/web3/SmartContractsClient.ts
```diff
@@ -271,7 +325,20 @@ export class SmartContractsClient
...
+   *
+   * @throws
+   * - If the contract binary data is missing.
+   * - If the contract contract address is missing.
+   * - If the result is empty.
+   * - If the result contains an error.
+   */
   public async executeReadOnlySmartContract(
     contractData: IContractData,
   ): Promise<IExecuteReadOnlyResponse> {
```
Ben-Rey, line 344: @AurelienFT Could you explain how execute_read_only_bytecode works ?

0xSwapFeeder (approved):

src/web3/SmartContractsClient.ts
Ben-Rey: @Leo-Besancon could you help us on that ?

Leo-Besancon: @Ben-Rey I'll probably have some time for you tomorrow, but I'm not sure how much help I can bring about this function ;)

Eitu33: @Ben-Rey Apologies, I missed the notification. A RO call or execution in massa has the same behaviour as a normal call or execution with the exception that every change made on the ledger by the given call / execution will get reverted after completion. Also another important thing to note is that a RO request will not create an operation (like CallSC or ExecuteSC would), it is considered as low priority execution, since it emits no operation it has no trace in blocks, and only one RO request can be executed per slot. Let me know if that answers your question.

Ben-Rey (approved):

PR #179, opened 2023-05-03
152 update event poller.ts documentation

Elli610:
related to #152

### DIFF
--- a/src/web3/EventPoller.ts
+++ b/src/web3/EventPoller.ts
@@ -25,7 +25,7 @@ const sortByThreadAndPeriod = (a: ISlot, b: ISlot): number => {
   return periodOrder;
 };
 
-/** 
+/**
  * The EventPoller object allows you to easily poll for events from the Massa network.
  */
 export class EventPoller extends EventEmitter {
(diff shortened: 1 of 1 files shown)


Ben-Rey (approved):

src/web3/EventPoller.ts
```diff
@@ -10,6 +10,14 @@ import { Client } from './Client';
 export const ON_MASSA_EVENT_DATA = 'ON_MASSA_EVENT';
 export const ON_MASSA_EVENT_ERROR = 'ON_MASSA_ERROR';
 
+/**
+ * Sorts slots by thread and period
```
gregLibert, line 14: This function us not sorting but comparing slot. If you agree, can you create an issue to fix it's name ?

gregLibert (approved):
I'm ok, except for the function name. I don't want to block this issue so i'm approving it. But please consider my comment

src/web3/EventPoller.ts
Elli610: issue created

Elli610:
> I'm ok, except for the function name. I don't want to block this issue so i'm approving it. But please consider my comment

I opened an issue about that

PR #194, opened 2023-05-03
[feat] add max_coins field

Ben-Rey:


examples/smartContracts/index.ts
```diff
@@ -211,6 +211,7 @@ const pollAsyncEvents = async (
         },
       ],
       web3Client,
+      fromMAS(0.1),
```
Ben-Rey, line 214: @AurelienFT For a deploy sc, does the max_coins will always be equal to the **coins: fromMAS(0.1),** above

AurelienFT: I don't think so because it depends on the storage costs it could use and the more the SC is bigger the more it will needs coins

Ben-Rey: ok so just to be sure, it's the sc builder responsibility to define max_coins

gregLibert (approved):

PR #205, opened 2023-05-10
remove the useless is_in_blockclique from IBlockInfo.ts

### LINKED ISSUE #201: The "is_in_blockclique" element  appears twice in "IBlockInfo.ts"
The one in content should be removed

Elli610:
closes #201

### DIFF
--- a/src/interfaces/IBlockInfo.ts
+++ b/src/interfaces/IBlockInfo.ts
@@ -6,7 +6,6 @@ export interface IBlockInfo {
   content: null | {
     is_final: boolean;
     is_stale: boolean;
-    is_in_blockclique: boolean;
     block: {
       header: {
         content: {


src/interfaces/IBlockInfo.ts
```diff
@@ -6,7 +6,6 @@ export interface IBlockInfo {
   content: null | {
     is_final: boolean;
     is_stale: boolean;
-    is_in_blockclique: boolean;
```
Ben-Rey, line 9: Can you share more information on why we need o remove it ?

Elli610: The is_in_blockClique appears twice : the first is in the "content" and the second one in the "root". I asked @AurelienFT if it was wanted and he told me that it was an "is_in_blockclique" located in the content was an error

Ben-Rey: great thanks!

Ben-Rey (approved):

PR #206, opened 2023-05-10
Update EventPoller.ts

### LINKED ISSUE #203: change the "sortByThreadAndPeriod " function name from EventPoller.ts
This function is not sorting but comparing slot

Elli610:
sortByThreadAndPeriod -> compareThreadAndPeriod

closes #203

### DIFF
--- a/src/web3/EventPoller.ts
+++ b/src/web3/EventPoller.ts
@@ -11,14 +11,14 @@ export const ON_MASSA_EVENT_DATA = 'ON_MASSA_EVENT';
 export const ON_MASSA_EVENT_ERROR = 'ON_MASSA_ERROR';
 
 /**
- * Sorts slots by thread and period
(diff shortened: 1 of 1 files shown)


Ben-Rey (changes requested):

Ben-Rey:
Suggestion
```
/**
 * Compares two ISlot instances based on their 'period' and 'thread' properties.
 *
 * @remarks
 * The comparison is primarily based on the 'period' property.  If the 'period' values are the same, 
 * the comparison is then based on the 'thread' property.
 * 
 * @param a - The first ISlot instance to be compared.
 * @param b - The second ISlot instance to be compared.
 *
 * @returns A positive number if 'a' should come after 'b', a negative number if 'a' should come before 'b',
 * or 0 if 'a' and 'b' are considered equal. 
 */
 ```

PR #210, opened 2023-05-15
156 check and update if necessary the interfaces documentation i private api client (included)   ws rpc methods (included) 

### LINKED ISSUE #156: Check and update if necessary the interfaces documentation : IPrivateApiClient (included) -> WsRpcMethods (included)

Elli610:
closes #156

### DIFF
--- a/src/interfaces/IRollsData.ts
+++ b/src/interfaces/IRollsData.ts
@@ -2,7 +2,7 @@
  * Represents the data of a roll.
  *
  * @see fee - fees paid for the roll
- * @see amount - amount paid for the roll
+ * @see amount - amount of rolls
  */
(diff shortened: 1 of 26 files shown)


src/interfaces/IRollsData.ts
```diff
@@ -1,3 +1,9 @@
+/**
+ * Represents the data of a roll.
+ *
+ * @see fee - fees paid for the roll
+ * @see amount - amount paid for the roll
```
Ben-Rey, line 5: Can you confirm it's not the amount of rolls ?

Elli610: you are right

PR #231, opened 2023-05-24
230 remove typescript any

### LINKED ISSUE #230: Remove typescript `any`

julienbrs:
closes #230

### DIFF
--- a/examples/smartContracts/deployer.ts
+++ b/examples/smartContracts/deployer.ts
@@ -79,7 +79,12 @@ export const deploySmartContracts = async (
   try {
     // do checks
     if (!deployerAccount) {
-      deployerAccount = web3Client.wallet().getBaseAccount()!;
+      const baseAccount = web3Client.wallet().getBaseAccount();
+      if (baseAccount === null) {
+        throw new Error('Failed to get base account');
+      } else {
+        deployerAccount = baseAccount;
+      }
     }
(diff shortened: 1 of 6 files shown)


examples/smartContracts/deployer.ts
```diff
@@ -79,7 +79,12 @@ export const deploySmartContracts = async (
   try {
     // do checks
     if (!deployerAccount) {
-      deployerAccount = web3Client.wallet().getBaseAccount()!;
+      const baseAccount = web3Client.wallet().getBaseAccount();
+      if (baseAccount === null) {
```
0xSwapFeeder, line 83: Maybe a ternary for if/else pattern to simplify, or even better since it throws:
an assertion ?
e.g:
```ts
deployerAccount = web3Client.wallet().getBaseAccount()
assert(deployerAccount !== null, "Failed to use base account");
```

0xSwapFeeder (changes requested):
Simple clarification request, otherwise lgtm.

examples/smartContracts/deployer.ts
0xSwapFeeder: Oh yeah you are right, forgot how dumb eslint was, you can mark as resolved then.

0xSwapFeeder (approved):

PR #253, opened 2023-06-02
remove fee from IReadData

### LINKED ISSUE #252: The "fee" from IReadData is never used and we don't need it here

Elli610:
closes #252 

@AurelienFT Can we remove the "fee" ? I can't find any method in web3 that's using it.


AurelienFT:
yes we are not using this parameter for a read smart contract

Ben-Rey (approved):

PR #256, opened 2023-06-02
251 refactor code for testnet 23

### LINKED ISSUE #251: refactor code for testnet 23
Cleaning the new code introduced with testnet 23, especially VERSION_NUMBER

julienbrs:
closes #251


src/utils/bytes.ts
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * @returns a Uint8Array containing the bytes of the secret key.
+ */
+export function getBytesSecretKey(secretKey: string): Uint8Array {
+  if (!(secretKey[0] == SECRET_KEY_PREFIX)) {
+    throw new Error(
+      `Invalid secret key prefix: "${secretKey[0]}". The secret key should start with "${SECRET_KEY_PREFIX}". Please verify your secret key and try again.`,
+    );
+  }
+  const secretKeyVersionBase58Decoded: Buffer = base58Decode(
```
julienbrs, line 21: Temporary function, will be removed in the future

julienbrs, line 21: Temporary function, will be deleted in the future

src/utils/keyAndAdresses.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+}
+
+export class PublicKey {
+  version: number;
+  publicKey: Uint8Array;
+  base58Encoded: string;
+
+  constructor(publicKeyArray: Uint8Array, secretKey: SecretKey) {
+    this.version = secretKey.version;
```
julienbrs, line 41: By creating the public key from the secret key object, it requires to have the same version for the public key and secret key. Thus, we can't have different versions.
We could do a different constructor with in arg the publicKeyBase58encoded, and so we could decode the version, but sometimes we don't have the publicKeyBase58encoded yet

src/web3/WalletClient.ts
```diff
@@ -328,31 +313,34 @@ export class WalletClient extends BaseClient implements IWalletClient {
 
   /**
    * Generates a new wallet account.
+   * @param version_number - The version number of the secret key to be generated, to create a new account.
    *
    * @returns A Promise that resolves to an {@link IAccount} object, which represents the newly created account.
    */
-  public static async walletGenerateNewAccount(): Promise<IAccount> {
+  public static async walletGenerateNewAccount(
+    version_number: number,
```
julienbrs, line 321: version number in arg to generate a new Account. Moreover, it is here the secret key version, and not the public key or address version

src/web3/WalletClient.ts
```diff
@@ -510,7 +489,7 @@ export class WalletClient extends BaseClient implements IWalletClient {
     }
 
     // convert signature to base58
-    const version = Buffer.from(varintEncode(VERSION_NUMBER));
+    const version = Buffer.from(varintEncode(VERSION_NUMBER)); // issue here
```
julienbrs, line 492: Will be replaced by: `const version = Buffer.from(varintEncode(secretKey.version));`
Signature version will be secretKey version, or we should put sig version in arg of the function

julienbrs: @AurelienFT

Ben-Rey (changes requested):

src/utils/keyAndAdresses.ts
```diff
@@ -0,0 +1,114 @@
...
+ * @remarks the public key is derived from the secret key and got the same version as the secret key.
+ */
+export class PublicKey {
+  version: number;
+  publicKey: Uint8Array;
+  base58Encoded: string;
+  base58Decoded: Uint8Array;
+  base58DecodedUnversioned: Uint8Array;
+
+  constructor(publicKeyArray: Uint8Array, secretKey: SecretKey) {
```
gregLibert, line 67: If the public key is derived fro the privatekey, why do you set a publickeyarray argument ?

src/utils/keyAndAdresses.ts
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * A secret key.
+ *
+ * @remarks the secret key object is created from a base58 encoded string representing the secret key.
+ * The secret key bytes representation is `version + secretKey` where secretKey is the 32 bytes of the secret key.
+ */
+export class SecretKey {
+  base58Encoded: string;
+  base58Decoded: Uint8Array;
+  base58DecodedUnversioned: Uint8Array;
```
gregLibert, line 29: What is behing this concept of base58DecodedUnversioned? Because by default base58 has no version, right?

src/utils/keyAndAdresses.ts
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * A secret key.
+ *
+ * @remarks the secret key object is created from a base58 encoded string representing the secret key.
+ * The secret key bytes representation is `version + secretKey` where secretKey is the 32 bytes of the secret key.
+ */
+export class SecretKey {
+  base58Encoded: string;
+  base58Decoded: Uint8Array;
```
gregLibert, line 28: base58 is an encoding format, like base64, which as nothing to do with the format that massa has chosen. So base58Decoded is a byte array, right ?

src/utils/keyAndAdresses.ts
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * @remarks the address object is created from a public key.
+ *
+ * @remarks the address bytes representation is `version + hashBlake3(version + publicKey)`.
+ *
+ * @remarks the address is derived from the public key and got the same version as the public key.
+ */
+export class Address {
+  prefix: string;
+  version: number;
```
gregLibert, line 94: is this the version of the address or of the public key ?

src/utils/keyAndAdresses.ts
```diff
@@ -0,0 +1,114 @@
...
+ * An address.
+ *
+ * @remarks the address object is created from a public key.
+ *
+ * @remarks the address bytes representation is `version + hashBlake3(version + publicKey)`.
+ *
+ * @remarks the address is derived from the public key and got the same version as the public key.
+ */
+export class Address {
+  prefix: string;
```
gregLibert, line 93: is this an attribute of the address object or simply a constant ?

src/web3/WalletClient.ts
```diff
@@ -236,42 +223,35 @@ export class WalletClient extends BaseClient implements IWalletClient {
...
-        secretKeyBase58Encoded,
-      );
-      // get public key
-      const publicKey: Uint8Array = await ed.getPublicKey(
-        secretKeyBase58Decoded,
+      const secretKey: SecretKey = new SecretKey(secretKeyBase58Encoded);
+
+      // create the public key object
+      const publicKeyArray: Uint8Array = await ed.getPublicKey(
+        secretKey.base58DecodedUnversioned,
```
gregLibert, line 232: This attribute is really unclear.

src/web3/WalletClient.ts
```diff
@@ -328,31 +308,34 @@ export class WalletClient extends BaseClient implements IWalletClient {
 
   /**
    * Generates a new wallet account.
+   * @param version_number - The version number of the secret key to be generated, to create a new account.
    *
    * @returns A Promise that resolves to an {@link IAccount} object, which represents the newly created account.
    */
-  public static async walletGenerateNewAccount(): Promise<IAccount> {
+  public static async walletGenerateNewAccount(
+    version_number: number,
```
gregLibert, line 316: Why is it an argument of the wallet generation function ? Can a user choose between version ?

gregLibert (changes requested):

src/utils/keyAndAdresses.ts
julienbrs: Here is the process to create a publicKey object:
```
const publicKeyArray: Uint8Array = await ed.getPublicKey(base58DecodedUnversioned);
const publicKey: PublicKey = new PublicKey(publicKeyArray, secretKey);
```
We derive the publicKey array bytes with an async function, that's why I didn't want to put it in the constructor. The secretKey arg in the constructor is only here to give the right version.
Maybe we should make a custom async constructor and put everything in it ? Thus there would be only the secretKey argument.

src/utils/keyAndAdresses.ts
julienbrs: Indeed attributes names aren't clear at all, I'll change that

src/web3/WalletClient.ts
julienbrs: I didn't want to hardcode the VERSION_NUMBER for maintainability reasons. However when generating a new wallet, there is no information available within the code to determine the desired version, so it needs to be specified as an argument.
In the future we should make this argument optional and fetch the current version if not provided (but then this would pose a problem for offline wallet generation)

Ben-Rey (approved):

src/utils/keyAndAdresses.ts
gregLibert: An async constructor feel weird. Can you put it in a FromPrivateKey function instead?
I still don't understand why the publicKeyArray is needed

src/web3/WalletClient.ts
gregLibert: I think you shouldn't let the choice to the caller and use the last known version instead

gregLibert (changes requested):
See my rseponse in the dicsussions

src/utils/keyAndAdresses.ts
```diff
@@ -66,24 +66,30 @@ export class SecretKey {
  */
 export class PublicKey {
   version: number;
-  publicKey: Uint8Array;
   base58Encoded: string;
   base58Decoded: Uint8Array;
   bytesUnversioned: Uint8Array;
 
-  constructor(publicKeyArray: Uint8Array, secretKey: SecretKey) {
+  private constructor(publicKeyArray: Uint8Array, secretKey: SecretKey) {
```
gregLibert, line 73:
```suggestion
private constructor(bytes: Uint8Array, version: number) {

File: src/utils/keyAndAdresses.ts [hunk 13]
```diff
@@ -66,24 +66,30 @@ export class SecretKey {
...
     this.version = secretKey.version;
-    this.publicKey = publicKeyArray;
+    this.bytesUnversioned = publicKeyArray;
```
gregLibert, line 75:
```suggestion
    this.bytes = bytes;
```

src/utils/keyAndAdresses.ts
```diff
@@ -66,24 +66,30 @@ export class SecretKey {
...
     const versionBuffer = Buffer.from(varintEncode(this.version));
 
     // Generate base58 encoded public key
     this.base58Encoded =
       PUBLIC_KEY_PREFIX +
-      base58Encode(Buffer.concat([versionBuffer, Buffer.from(this.publicKey)]));
+      base58Encode(
+        Buffer.concat([versionBuffer, Buffer.from(this.bytesUnversioned)]),
+      );
     this.base58Decoded = getBytesPublicKey(this.base58Encoded);
```
gregLibert:
```suggestion
this.base58EncodeString = PUBLIC_KEY_PREFIX +
      base58Encode(
        Buffer.concat([versionBuffer, Buffer.from(this.bytesUnversioned)]),
      );

this.base58EncodeBytes = getBytesPublicKey(this.base58EncodeString);
```

src/utils/keyAndAdresses.ts
```diff
@@ -66,24 +66,30 @@ export class SecretKey {
...
+  }
 
-    // Slice off the version byte
-    this.bytesUnversioned = this.base58Decoded.slice(1);
+  // Call this method to create a new PublicKey from a secret key.
+  static async fromSecretKey(secretKey: SecretKey): Promise<PublicKey> {
+    const publicKeyArray: Uint8Array = await ed.getPublicKey(
```
gregLibert, line 89:
```suggestion
    const bytes: Uint8Array = await ed.getPublicKey(
```

src/utils/keyAndAdresses.ts
```diff
@@ -66,24 +66,30 @@ export class SecretKey {
...
+      secretKey.bytesUnversioned,
+    );
+    return new PublicKey(publicKeyArray, secretKey);
```
gregLibert, line 92: You don't need to send the privateKey here, only it versoin

src/utils/keyAndAdresses.ts
```diff
@@ -27,14 +28,12 @@ const ADDRESS_PREFIX = 'AU';
  * @remarks bytesUnversioned attribute is the Uint8Array representation of the public key without the version byte.
  */
 export class SecretKey {
+  version: number;
   base58Encoded: string;
   base58Decoded: Uint8Array;
   bytesUnversioned: Uint8Array;
```
gregLibert:
```suggestion
  base58EncodeString: string;
  base58EncodeBytes: Uint8Array;
  bytes: Uint8Array;
```

gregLibert (changes requested):

src/utils/keyAndAdresses.ts
```diff
@@ -0,0 +1,121 @@
...
+    this.base58EncodeBytes = getBytesSecretKey(secretKeyBase58Encoded);
+
+    // Slice off the version byte
+    this.bytes = this.base58EncodeBytes.slice(1);
+
+    const secretKeyUnprefixed = secretKeyBase58Encoded.slice(
+      SECRET_KEY_PREFIX.length,
+    );
+    const decoded = base58Decode(secretKeyUnprefixed);
+    const decodedVersion = varintDecode(decoded);
```
gregLibert, line 47: already done in getBytesSecretKey

src/utils/keyAndAdresses.ts
```diff
@@ -0,0 +1,121 @@
...
+ */
+export class SecretKey {
+  version: number;
+  base58EncodeString: string;
+  base58EncodeBytes: Uint8Array;
+  bytes: Uint8Array;
+
+  constructor(secretKeyBase58Encoded: string) {
+    this.base58EncodeString = secretKeyBase58Encoded;
+    this.base58EncodeBytes = getBytesSecretKey(secretKeyBase58Encoded);
```
gregLibert, line 38:
```suggestion
    versionAndKeyBytes = getBytesSecretKey(secretKeyBase58Encoded);
```

src/utils/keyAndAdresses.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    // Slice off the version byte
+    this.bytes = this.base58EncodeBytes.slice(1);
+
```
gregLibert, line 42:
```suggestion
   this.version = varintDecode(versionAndKeyBytes[0])
```

src/utils/keyAndAdresses.ts
```diff
@@ -0,0 +1,121 @@
...
+ * The secret key object is created from a base58 encoded string representing the secret key.
+ * The secret key bytes representation is `version + secretKey` where secretKey is the 32 bytes of the secret key.
+ *
+ * @remarks base58Encoded attribute is the readable string representation of the public key.
+ * @remarks base58Decoded attribute is the Uint8Array representation of the public key.
+ * @remarks bytesUnversioned attribute is the Uint8Array representation of the public key without the version byte.
+ */
+export class SecretKey {
+  version: number;
+  base58EncodeString: string;
```
gregLibert, line 32: private

src/utils/keyAndAdresses.ts
```diff
@@ -0,0 +1,121 @@
...
+  base58EncodeBytes: Uint8Array;
+  bytes: Uint8Array;
```
gregLibert, line 34: private

src/utils/keyAndAdresses.ts
```diff
@@ -0,0 +1,121 @@
...
+    const secretKeyUnprefixed = secretKeyBase58Encoded.slice(
+      SECRET_KEY_PREFIX.length,
+    );
+    const decoded = base58Decode(secretKeyUnprefixed);
+    const decodedVersion = varintDecode(decoded);
+
+    // Extract version from the secret key
+    this.version = decodedVersion.value;
+  }
+}
```
gregLibert, line 52: add getPublickKey()

src/utils/keyAndAdresses.ts
```diff
@@ -0,0 +1,121 @@
...
+ * The secret key bytes representation is `version + secretKey` where secretKey is the 32 bytes of the secret key.
+ *
+ * @remarks base58Encoded attribute is the readable string representation of the public key.
+ * @remarks base58Decoded attribute is the Uint8Array representation of the public key.
+ * @remarks bytesUnversioned attribute is the Uint8Array representation of the public key without the version byte.
+ */
+export class SecretKey {
+  version: number;
+  base58EncodeString: string;
+  base58EncodeBytes: Uint8Array;
```
gregLibert, line 33: to be suppressed

src/utils/keyAndAdresses.ts
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * @remarks the address object is created from a public key.
+ *
+ * @remarks the address bytes representation is `version + hashBlake3(version + publicKey)`.
+ *
+ * @remarks the address is derived from the public key and got the same version as the public key.
+ */
+export class Address {
+  version: number;
+  base58EncodeString: string;
```
gregLibert, line 103:
```suggestion
  base58Encode: string;
```

gregLibert (changes requested):

src/utils/keyAndAdresses.ts
```diff
@@ -1,121 +1,99 @@
...
+    const publicKeyArray: Uint8Array = await ed.getPublicKey(this.bytes);
+    return new PublicKey(publicKeyArray, this.version);
+  }
 
-    // Extract version from the secret key
-    this.version = decodedVersion.value;
+  /* Sign a message hash digest with the secret key */
+  async signDigest(messageHashDigest: Uint8Array): Promise<Uint8Array> {
+    const sig = await ed.sign(messageHashDigest, this.bytes);
+    return sig;
```
gregLibert, line 46:
```suggestion
    return await ed.sign(messageHashDigest, this.bytes);
```

gregLibert (changes requested):
It's cleaner, well done!
Only one small code suggestion left

PR #277, opened 2023-06-13
273 fix ci

gregLibert:


gregLibert:
I'm stuck: we need to run the test on two different networks: buildnet and testnet. The only ways I found to solve this are:
- remove the test form here (they are not unit test) and move them in the proper end to end framework
- have two different branches to handle two different versions

Obviously option is cleaner...

julienbrs:
> I'm stuck: we need to run the test on two different networks: buildnet and testnet. The only ways I found to solve this are:
> 
>     * remove the test form here (they are not unit test) and move them in the proper end to end framework
> 
>     * have two different branches to handle two different versions
> 
> 
> Obviously option is cleaner...

I think we won't need these test examples anymore giving the unit tests to come. Moreover, is is perhaps planned to have a different repo with integration tests (like the ones here in the CI) to ensure massa-web3 is working with different frameworks (vite, next,...) that could be launched manually from time to time

PR #285, opened 2023-06-15
284 tests baseclient setProviders, getPrivateProviders & getPublicProviders

julienbrs:
At the moment we can set several publicProviders or several privateProviders at the same time, is it something that we want ? 
Moreover, most of the functions tested here are `protected`, so we need to put them `public` to test it. Idk if it's a problem

### DIFF
--- a/src/web3/BaseClient.ts
+++ b/src/web3/BaseClient.ts
@@ -91,6 +91,25 @@ export class BaseClient {
    * @param providers - The new providers to set as an array of IProvider.
    */
   public setProviders(providers: Array<IProvider>): void {
+    const hasPublicProvider = providers.some(
+      (provider) => provider.type === ProviderType.PUBLIC,
+    );
+    const hasPrivateProvider = providers.some(
+      (provider) => provider.type === ProviderType.PRIVATE,
+    );
+
+    if (!hasPublicProvider) {
+      throw new Error(
+        'Cannot set providers with no public providers. Need at least one',
+      );
+    }
+
+    if (!hasPrivateProvider) {
+      throw new Error(
+        'Cannot set providers with no private providers. Need at least one',
+      );
+    }
+
     this.clientConfig.providers = providers;
   }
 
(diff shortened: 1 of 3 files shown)


src/web3/BaseClient.ts
```diff
@@ -99,7 +118,7 @@ export class BaseClient {
    *
    * @returns An array of IProvider containing all the private providers.
    */
-  protected getPrivateProviders(): Array<IProvider> {
+  public getPrivateProviders(): Array<IProvider> {
```
Ben-Rey, line 121: I don't think making this function public is the best choice. Protected functions protect us from breaking changes.
I suggest 2 ideas:
1. Through the constructor
2. Test subclass: A subclass specifically for testing, which exposes the protected functions.

src/web3/BaseClient.ts
```diff
@@ -91,6 +91,25 @@ export class BaseClient {
    * @param providers - The new providers to set as an array of IProvider.
    */
```
Ben-Rey, line 92: Could you add throw to the typedoc ?

Ben-Rey (changes requested):

Ben-Rey (approved):

test/web3/baseClient.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+import { IProvider, ProviderType } from '../../src/interfaces/IProvider';
+import { IClientConfig } from '../../src/interfaces/IClientConfig';
+
+// for CI testing:
+const publicApi = 'http://mock-public-api.com';
+const privateApi = 'http://mock-private-api.com';
+
+// For local testing:
+// const publicApi = 'http://127.0.0.1:33035';
+// const privateApi = 'http://127.0.0.1:33034';
```
gregLibert, line 11:
```suggestion
const publicApi = process.env.CI ? 'http://mock-public-api.com' : 'http://127.0.0.1:33035';
const privateApi = process.env.CI ? 'http://mock-private-api.com' : 'http://127.0.0.1:33034';
```

gregLibert (changes requested):

PR #301, opened 2023-06-19
297 tests eventPoller:  startPolling, stopPolling & compareByThreadAndPeriod

julienbrs:


julienbrs:
@gregLibert can we merge even if SonarCloud detects code duplication? It's not actually code duplication, but rather some duplicated test assertions.

gregLibert (approved):

PR #311, opened 2023-06-20
Tests walletclient sendtransaction

julienbrs:


test/web3/walletClient.spec.ts
```diff
@@ -15,12 +19,12 @@ const receiverPrivateKey =
...
-// const privateApi = 'https://test.massa.net/api/v2:33034';
+const publicApi = 'https://mock-public-api.com';
+const privateApi = 'https://mock-private-api.com';
 
-// For local testing:
-const publicApi = 'http://127.0.0.1:33035';
-const privateApi = 'http://127.0.0.1:33034';
+// // For local testing:
+// const publicApi = 'http://127.0.0.1:33035';
+// const privateApi = 'http://127.0.0.1:33034';
```
gregLibert, line 27: Can you please use the same technic as in the last PR that needed to have both local and CI configuration?

gregLibert (changes requested):

test/web3/walletClient.spec.ts
julienbrs: It's not necessary anymore, I'm mocking every call to the node. So these urls don't matter, i'll change it to mock url and remove the local testing url when every PR will be merged

gregLibert (approved):

PR #325, opened 2023-06-22
322 test smartContractsClient: deploySmartcontract

julienbrs:


src/web3/SmartContractsClient.ts
```diff
@@ -97,6 +97,11 @@ export class SmartContractsClient
     const expiryPeriod: number =
       nodeStatusInfo.next_slot.period + this.clientConfig.periodOffset;
 
+    // Check if SC data exists
+    if (!contractData.contractDataBinary) {
+      throw new Error(`Contract data required. Got null`);
```
gregLibert, line 102:
```suggestion
      throw new Error(`Expected non-null contract bytecode, but received null.`);

```

gregLibert (changes requested):

gregLibert (approved):

PR #362, opened 2023-06-27
360 mock tests getaccountbalance + small additional tests

julienbrs:
- mock getAddresses calls in tests of getAccountBalance so we can run tests without to call a node
- add a test for walletSignMessage to check that we throw an error when the signature could not be verified


Elli610 (approved):
looks good to me

gregLibert (approved):

PR #372, opened 2023-06-28
add buildnet in defaultProvider

Ben-Rey:


julienbrs (approved):
We should also add `BUILDNET` and `LOCALNET` to `IClient.ts`, `

PR #402, opened 2023-07-13
add first test end to end wallet

### LINKED ISSUE #398: test end-to-end wallet
End-to-end test that need to have all the functions mentioned in the `massa-web3` page in `docu-dev`, at section `Wallet` (so all functions here https://web3.docs.massa.net/classes/WalletClient.html).
Existing example at `examples/wallet` will be adapted for that.

julienbrs:
closes #398


gregLibert:
I think that this is not the right direction.

I think that we should:
- Create a real front app using the functionalities we want to test.
- Create end user tests using something like Cypress or Puppeteer to check that the application is working as expected.
- Execute the tests in the CI.

@Ben-Rey Wdyt ?
@mazmassa Do you have any advice to share on the test framework to use for end user tests ?

mazmassa:
> I think that this is not the right direction.
> 
> I think that we should:
> 
> * Create a real front app using the functionalities we want to test.
> * Create end user tests using something like Cypress or Puppeteer to check that the application is working as expected.
> * Execute the tests in the CI.
> 
> @Ben-Rey Wdyt ? @mazmassa Do you have any advice to share on the test framework to use for end user tests ?

Speaking about the wallet we should indeed go through Cypress for the acceptance ones (e2e). Starting by the happy path test then all variations we can have in the user world. However, about the massaweb3 I would not go with acceptance ones but in unit/integration ones.

julienbrs:
Closing this PR, e2e test will be implemented for the interesting examples that are in `massa-sc-examples` with cypress and a correct CI, we don't need a specific test for `massa-web3` like this one anymore.

PR #423, opened 2023-07-26
Introduce wallet provider

AurelienFT:
This PR : 

- Remove the limit to have a provider of private API.
- Use random provider to make an API call so that if we pass a big list from wallet-provider it will switch between nodes often.
- Abstract low level operation creation in 2 mode : One with local secret key, one using wallet provider
- Add a constructor of client that take account of wallet provider as parameters.

This PR is used in an example to test it : https://github.com/massalabs/massa-sc-examples/pull/157


AurelienFT:
@gregLibert Can we ignore the security concerns as it's not important in our use case ?

src/web3/Client.ts
```diff
@@ -26,15 +33,30 @@ export class Client implements IClient {
...
+    baseAccount?: IAccount | IAccountWalletProvider,
   ) {
     this.publicApiClient = new PublicApiClient(clientConfig);
     this.privateApiClient = new PrivateApiClient(clientConfig);
-    this.walletClient = new WalletClient(
-      clientConfig,
-      this.publicApiClient,
-      baseAccount,
-    );
+    // From: https://stackoverflow.com/a/73162923/10419533
```
gregLibert, line 40: Not sure it's linked to what you are doing.
Here your simply defining the base account based on the wanted context. If you don't want to go the direction I suggested above, you could improve readibility by doing something like:
``` typescript
this.walletClient = new WalletClient(
        clientConfig,
        this.publicApiClient,
        provider(baseAccount, this.publicApiClient),
      );
 ```
 And somewhere else define the provider function:
 ```typescript
  private provider(): Account {
  if (!baseAccount) {
    return undefined;
  }

  if (isWeb3Account(baseAccount)) {
    return new Web3Account(baseAccount, this.publicApiClient);
  }
  
  return new WalletProviderAccount(baseAccount);
  }

File: src/web3/Client.ts [hunk 2]
```diff
@@ -26,15 +33,30 @@ export class Client implements IClient {
    */
   public constructor(
     private clientConfig: IClientConfig,
-    baseAccount?: IAccount,
+    baseAccount?: IAccount | IAccountWalletProvider,
```
gregLibert, line 36: Why not using a IBaseAccount instead and let people instantiate Client class using meaningful helper such as:
const client =  newBrowserWalletProvider() or const client = newNodeJSProvider() ?
this abviously imply to define the function newBrowserWalletProvider and newNodeJSProvider

gregLibert (commented):
To be discussed in live meeting

src/web3/BaseClient.ts
```diff
@@ -185,7 +173,9 @@ export class BaseClient {
       case JSON_RPC_REQUEST_METHOD.ADD_STAKING_PRIVATE_KEYS:
       case JSON_RPC_REQUEST_METHOD.NODE_SIGN_MESSAGE:
       case JSON_RPC_REQUEST_METHOD.NODE_REMOVE_FROM_WHITELIST: {
-        return this.getPrivateProviders()[0]; // TODO: choose the first available private provider ?
+        let providers = this.getPrivateProviders();
```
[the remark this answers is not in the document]
AurelienFT: Yes but it's not mandatory anymore.

src/web3/Client.ts
```diff
@@ -7,7 +7,14 @@ import { SmartContractsClient } from './SmartContractsClient';
 import { IProvider, ProviderType } from '../interfaces/IProvider';
 import { DefaultProviderUrls } from './ClientFactory';
 import { IClient } from '../interfaces/IClient';
-import { IWalletClient } from '../interfaces/IWalletClient';
+import { IAccount as IAccountWalletProvider } from '@massalabs/wallet-provider';
+import { Web3Account } from './accounts/Web3Account';
+import { WalletProviderAccount } from './accounts/WalletProviderAccount';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function isWeb3Account(item: any): item is IAccount {
```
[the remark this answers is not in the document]
AurelienFT: Removed

src/web3/Client.ts
AurelienFT: Changed the behavior to have IBaseAccount

src/web3/Client.ts
AurelienFT: Done

Ben-Rey (approved):

gregLibert (approved):

PR #467, opened 2023-08-29
add validateAddress

### LINKED ISSUE #466: allow address validation

Elli610:
closes #466

### DIFF
--- a/src/web3/WalletClient.ts
+++ b/src/web3/WalletClient.ts
@@ -9,6 +9,7 @@ import {
   base58Encode,
   varintEncode,
   hashBlake3,
+  varintDecode,
 } from '../utils/Xbqcrypto';
 import { JSON_RPC_REQUEST_METHOD } from '../interfaces/JsonRpcMethods';
 import { trySafeExecute } from '../utils/retryExecuteFunction';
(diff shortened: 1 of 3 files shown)


src/web3/WalletClient.ts
```diff
@@ -564,4 +569,25 @@ export class WalletClient extends BaseClient implements IWalletClient {
...
+
+  /**
+   * Validate an address.
+   *
+   * @param address - The address to validate.
+   *
+   * @returns A boolean (true if the address is valid, false otherwise).
+   */
+  validateAddress(address: string): boolean {
+    address = address.slice(ADDRESS_PREFIX.length); // remove The address prefix
```
Ben-Rey, line 581: Should we check for the right address prefix ?

Ben-Rey (changes requested):

src/web3/WalletClient.ts
```diff
@@ -564,4 +569,25 @@ export class WalletClient extends BaseClient implements IWalletClient {
...
+    // check if the address is base58 encoded
+    try {
+      const bytesAddress = new Uint8Array(base58Decode(address));
+      if (bytesAddress.length != 32) {
```
Ben-Rey, line 585: In the rust side they check if the length of the address, `without the version` is equal to 32

And the version is an U64 varint

@sydhds please correct me if I misunderstood something

Elli610: updated !

src/web3/WalletClient.ts
```diff
@@ -564,4 +570,28 @@ export class WalletClient extends BaseClient implements IWalletClient {
...
+   *
+   * @returns A boolean (true if the address is valid, false otherwise).
+   */
+  validateAddress(address: string): boolean {
+    address = address.slice(ADDRESS_PREFIX.length); // remove The address prefix
+    // check if the address is base58 encoded
+    try {
+      // remove the version byte
+      const versionAndKeyBytes = new Uint8Array(base58Decode(address));
+      varintDecode(versionAndKeyBytes.slice(0, 1)); // check the version byte
```
sydhds, line 587: it seems that the varintDecode is done only on 1 byte. In Rust code, the version is encoded as a u64 varint so I guess its size could be more than 1 byte no?

Ben-Rey:
should be done here https://github.com/massalabs/massa-web3/pull/470

PR #502, opened 2023-11-22
Bump wallet pro

Ben-Rey:


packages/massa-web3/src/web3/accounts/WalletProviderAccount.ts
```diff
@@ -60,15 +60,16 @@ export class WalletProviderAccount implements IBaseAccount {
...
+    let res = (await this.account.callSC(
       callData.targetAddress,
       callData.functionName,
       paramToSend,
       callData.coins,
       callData.fee,
       callData.maxGas,
       false,
-    );
+    )) as ITransactionDetails;
```
peterjah, line 71: why ? :(
`let res: ITransactionDetails = await .... ` should work. but it should be infered

peterjah (changes requested):

peterjah (approved):

PR #519, opened 2024-01-05
Add coin and fee parameters to IReadData interface

0xB19:


sydhds (approved):

packages/massa-web3/src/interfaces/ICallData.ts
```diff
@@ -5,10 +5,10 @@ import { Args } from '@massalabs/web3-utils';
  *
  * @remarks
  * This interface is used to track the calldata information, including the
- * storage fee, maximum gas, transferred coins, target smart contract address,
+ * transaction fee, maximum gas, transferred coins, target smart contract address,
```
Thykof, line 8: maybe the wording would be better by saying 'operation'?

Ben-Rey: I think it would be confusing to talk about operation fees, as this is a very general term. In the context of blockchain and smart contracts, the term "transaction fees" is more commonly used.

Thykof (approved):

PR #529, opened 2024-02-05
528 use read sc to set gas when gas is not set in call sc

thomas-senechal:


Ben-Rey:
Could you remove the maxGas parameter here (and maybe for all callSc) so we have a real life example ? https://github.com/massalabs/massa-web3/blob/3325ff5c4bf368ae4c6e353882b14abc2a8fa59f/packages/massa-web3/examples/smartContracts/index.ts#L289

packages/massa-web3/src/web3/SmartContractsClient.ts
```diff
@@ -147,7 +147,7 @@ export class SmartContractsClient
 
     if (callData.maxGas === null || callData.maxGas === undefined) {
       const reponse = await this.readSmartContract({
-        maxGas: BigInt(0),
+        maxGas: BigInt(2_100_000),
```
Ben-Rey, line 150: Could you add a comment to explain why we need this value and maybe make a constant in web3-utils if that make sense

packages/massa-web3/code-snippets/walletGenerateNewAccount.ts
```diff
@@ -1,13 +1,13 @@
...
-// print the account address, public and private keys
-console.log(
-  'Address: ',
-  newAccount.address,
-  'Public key',
-  newAccount.publicKey,
-  'Private key',
-  newAccount.secretKey
-)
+;(async () => {
```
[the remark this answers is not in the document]
Ben-Rey: I had the same thought but it's something added automatically. I found this https://stackoverflow.com/questions/7365172/semicolon-before-self-invoking-function

packages/massa-web3/src/web3/SmartContractsClient.ts
```diff
@@ -144,6 +145,19 @@ export class SmartContractsClient
...
     }
+
+    if (callData.maxGas === null || callData.maxGas === undefined) {
+      const reponse = await this.readSmartContract({
+        maxGas: BigInt(BASE_INSTANCE_CREATION_GAS_COST),
+        targetAddress: callData.targetAddress,
+        targetFunction: callData.targetFunction,
+        parameter: callData.parameter,
+        coins: callData.coins,
+        fee: callData.fee,
```
Thykof, line 156:
```suggestion
        ...callData,
        maxGas: BigInt(BASE_INSTANCE_CREATION_GAS_COST),
```

Thykof (approved):

Thykof (approved):

Thykof (approved):

Thykof:
I think it doesn't work because the coins parameter must be in MAS for the read only call and in nMAS for the actual call. So when you pass a value for coins in nMAS, the read only call take it a MAS, and so you don't have enough on your balance to pay the coins.

PR #531, opened 2024-02-13
528 2 convert coins to mas

Thykof:
Warning: breaking change in the read smart contract function, you must now put the value in nanoMASSA.

Note: I can't convert the coins parameter in the call smart contract function because `ICallData.coins` is `bigint`, so the value in MASSA with a coma won't work. This is why I convert to MASSA in the read smart contract function. We can change the type of `ICallData.coins` to string or BigNumber if we don't want this breaking change.

Here is what we can do:

```ts
      try {
        const previousCoins = callData.coins;
        callData.coins = BigInt(toMAS(callData.coins).toString()); // error here, can't create a BigInt for a string with decimal part
        const response = await this.readSmartContract(callData)
        callData.maxGas = BigInt(response.info.gas_cost)
        callData.coins = previousCoins;
      } catch (error) {
        throw new Error(
          `Operation failed: Max gas unspecified and auto-estimation failed. Error details: ${error.message}`
        )
      }
```

### DIFF
--- a/.lintstagedrc
+++ b/.lintstagedrc
@@ -1,5 +1,5 @@
 {
-    "*.{js,ts,json,cjs,md}": [
+    "*.{js,ts,json,cjs}": [
         "prettier --check",
         "eslint"
     ]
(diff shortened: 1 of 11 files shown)


Ben-Rey (approved):

Thykof (commented):
OK

Ben-Rey (approved):

PR #532, opened 2024-02-14
Improve/get status

Ben-Rey:


packages/massa-web3/src/web3/SmartContractsClient.ts
```diff
@@ -338,29 +347,39 @@ export class SmartContractsClient
    * @returns A promise that resolves to the status of the operation.
    */
   public async getOperationStatus(opId: string): Promise<EOperationStatus> {
-    const operationData: Array<IOperationData> =
-      await this.publicApiClient.getOperations([opId])
-
-    if (!operationData?.length) return EOperationStatus.NOT_FOUND
+    const [operation] = await this.publicApiClient.getOperations([opId])
```
Thykof, line 350: this will throw an error if `await this.publicApiClient.getOperations([opId])` return an empty array

packages/massa-web3/src/web3/SmartContractsClient.ts
```diff
@@ -405,6 +424,42 @@ export class SmartContractsClient
...
+        }
+
+        if (
+          newStatus === EOperationStatus.FINAL_SUCCESS ||
+          newStatus === EOperationStatus.FINAL_ERROR
+        ) {
+          clearInterval(interval)
+        }
+      } catch (error) {
+        console.warn(error)
```
Thykof, line 456: linter says `Unexpected console statement` maybe consider removing this rule if we deliberately want to console.log

Thykof (dismissed):

packages/massa-web3/src/web3/SmartContractsClient.ts
Ben-Rey: Thanks!

packages/massa-web3/src/web3/SmartContractsClient.ts
Ben-Rey: This is just a warning, and I think we should keep it for now, as it prevents us from forgetting debug purpose console.log.

Ben-Rey: Ok I understand you opinion. I didn't added the rule myself so I need to think about it and decide if we really want to use console here

Ben-Rey: I tried another approche that allow the user to have more control what do you think ?

Thykof (approved):

packages/massa-web3/src/web3/SmartContractsClient.ts
```diff
@@ -405,6 +426,52 @@ export class SmartContractsClient
...
+   * @param timeInterval - The interval in milliseconds at which to check the operation status. Defaults to `TX_POLL_INTERVAL_MS`.
+   * @param timeout - The time at which to stop checking the operation status. Defaults to the current time plus `WAIT_OPERATION_TIMEOUT`.
+   *
+   * @returns A Promise that resolves to a function that can be called to stop watching the operation status.
+   */
+  public async watchOperationStatus(
+    opId: string,
+    callback: (status: EOperationStatus, error?: Error) => void,
+    timeInterval = TX_POLL_INTERVAL_MS,
+    timeout = Date.now() + WAIT_OPERATION_TIMEOUT
```
peterjah, line 443: its strange to have `Date.now()` here. usually we set a timeout range (like 3 secondes) implicitely starting from now.

peterjah (approved):

packages/massa-web3/src/web3/SmartContractsClient.ts
Ben-Rey: I updated the code, I don't know if it's what you was saying

PR #550, opened 2024-04-08
Add first version experimental account

AurelienFT:


packages/massa-web3/src/experimental/account.ts
```diff
@@ -0,0 +1,342 @@
...
+}
+
+export class SecretKey {
+  private bytes: Uint8Array
+  public version: Version
+
+  private constructor(bytes: Uint8Array, version: Version) {
+    this.bytes = bytes
+    this.version = version
+  }
```
gregLibert, line 21: This is a duplicate of SecretKey.fromBytes. Furthermore, why not injecting the following dependencies instead:
- string format (encode, decode of base58 for instance)
- crypto functions (sign, generateKeys, hash)

the probability that we change the crypto primitives is low, but non null. injecting dependencies will make this modification smoother and will make unit test coverage easier.

packages/massa-web3/src/experimental/account.ts
```diff
@@ -0,0 +1,342 @@
...
+   * @param version - The version of the signature.
+   *
+   * @returns A new signature object.
+   */
+  public static fromBytes(bytes: Uint8Array, version: Version): Signature {
+    return new Signature(bytes, version)
+  }
+}
+
+export class Account {
```
gregLibert, line 271: What about separating Account from Accountv0, Accountv1... ? Accountv0 having a specific configuration in terms of:
- crypto
- string serialisation/deserialisation
- ...

gregLibert (commented):
My comments focus solely on the structuring of the objects and ignore the rest.

packages/massa-web3/src/experimental/account.ts
```diff
@@ -477,4 +493,94 @@ export class Account {
...
+  /**
+   * Decode a keystore object following the Massa standard format
+   * https://github.com/massalabs/massa-standards/blob/main/wallet/file-format.md
+   * @param keystore - The keystore object to decode.
+   * @param password - The password to unseal the private key.
+   *
+   * @returns A new account object.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  static async fromKeyStore(keystore: any, password: string): Promise<Account> {
```
gregLibert, line 561: keystore should reflect existing contraints:
- go with a generic type and define v0 keystore format
- define a minimalist keystore with a version key and key: value for the rest.

```typescript
interface KeyStore {
  [key: string]: {
    version: Version
  };
}

interface AccountKeyStoreV0 {
          address: string,
          version: Version,
          nickname: string,
          ...
}

  static async fromKeyStore<T extends KeyStore>(keystore: T, password: string): Promise<Account> {
  ...
```

gregLibert (approved):

PR #551, opened 2024-04-08
Add fee by default for readonly

AurelienFT:


Thykof (approved):

Thykof:
we reintroduce a regression with that:

```
{"jsonrpc":"2.0","method":"execute_read_only_call","params":[[{"max_gas":4294167295,"target_address":"AS1Y1zNtZRUruRuqwN47YaYiEdD8gEoXjrEKKnA19X1BFypRGEqM","target_function":"supportedTokensList","parameter":[],"fee":"0.000000000"}]],"id":0}
```

gives this error: `
readonly call failed: Runtime error: spending address AU1t1XykhgFoySuxsULQJPKcNTMZBXsKT8uqGQ3YNvTrt3Unbykh not found`

this is because there is 0 fee and no caller address in the parameters of the rpc call and so the node generate one randomly and it doesn't exist so impossible to transfer 0 fee from it.

By the ways transfer 0 from a non existant address, does it make sense?

To fix this regression: 
```ts
    let fee: string | undefined;
    if (readData.fee) {
      fee = toMAS(readData.fee).toFixed(MASSA_SCALING_FACTOR);
    } else {
      fee = minimalFees > 0n ? toMAS(minimalFees).toFixed(MASSA_SCALING_FACTOR) : undefined;
    }
```

Thykof:
no, we will rollback this PR and modify the massa-node

Thykof:
https://github.com/massalabs/massa-web3/pull/554

AurelienFT:
Modification on the node : https://github.com/massalabs/massa/pull/4675

PR #559, opened 2024-04-18
558 xp add operation and accoutmanager

gregLibert:


packages/massa-web3/src/experimental/accountOperation.ts
```diff
@@ -0,0 +1,102 @@
...
+ */
+export class AccountOperation {
+
+    constructor(public account: Account, public client: BlockchainClient) {}
+  
+    private async rollOperation( amount: number, opts?: OptOpDetails): Promise<string> {
+      const operation = new OperationManager(this.account.privateKey, this.client)
+      const details: RollOperation = {
+        // Todo: change with fetchMinimalFees once ready
+        fee: opts?.fee || 0,
```
peterjah, line 17: prefere the use of `??` operator here instead of `||`.  (even if the result will be the same in this case, its more coherent. )

packages/massa-web3/src/experimental/publicAPI.ts
```diff
@@ -193,12 +221,28 @@ export class PublicAPI {
     return this.connector.get_status()
   }
 
-  async sendOperation(data: OperationInput): Promise<OperationId> {
+  async fetchChainId(): Promise<number> {
+    return this.getStatus().then((r) => r.chain_id)
+  }
```
peterjah, line 226: Its a bit sad to make a rpc call each time we need a info from getStatus.
Lets see how it will be used, but imho, we should fetch the constant data of getStatus (like chainId)  once for all at initialization of the client.
But maybe its more an optimization

peterjah (approved):
Nice one! see some little suggestions in comments

packages/massa-web3/src/experimental/accountOperation.ts
```diff
@@ -0,0 +1,102 @@
...
+     * Transfers tokens.
+     * 
+     * @param to - The address of the recipient.
+     * @param amount - The amount of tokens to transfer.
+     * @param opts - Optional operation details.
+     * 
+     * @returns The ID of the operation.
+     * @throws If the amount of tokens is not a positive non-zero value.
+     */
+    async transfer(to: Address | string, amount: number, opts?: OptOpDetails): Promise<Operation> {
```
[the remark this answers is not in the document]
gregLibert: No, amount, gas, fee,... should be bigint.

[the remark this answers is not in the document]
gregLibert: I propose to address this problem once all the PR will be merge.

packages/massa-web3/src/experimental/basicElements/address.ts
```diff
@@ -140,6 +155,6 @@ export class Address {
   toString(): string {
     return `${ADDRESS_PREFIX}${
       this.isEOA ? ADDRESS_USER_PREFIX : ADDRESS_CONTRACT_PREFIX
-    }${this.serializer.serialize(this.versionedBytes())}`
+    }${this.serializer.serialize(this.versionedBytes().slice(1))}`
```
AurelienFT, line 158: This is unclear but I don't find ways to improve this right now in my head.

packages/massa-web3/src/experimental/accountOperation.ts
```diff
@@ -0,0 +1,133 @@
...
+  /**
+   * Fetches the balance of the account.
+   *
+   * @param speculative - Whether to fetch the speculative balance. Default is the final balance.
+   *
+   * @returns The balance of the account.
+   */
+  async fetchBalance(speculative: boolean = false): Promise<number> {
+    return this.client.getBalance(this.account.address.toString(), speculative)
+  }
```
AurelienFT, line 132: is it expected to not include deploy sc, call sc, read only call and estimate gas call ?

packages/massa-web3/src/experimental/accountOperation.ts
```diff
@@ -0,0 +1,133 @@
...
+   * @param opts - Optional operation details.
+   *
+   * @returns The ID of the operation.
+   * @throws If the amount of rolls is not a positive non-zero value.
+   */
+  async buyRolls(
+    client: BlockchainClient,
+    amount: number,
+    opts?: OptOpDetails
+  ): Promise<string> {
```
AurelienFT, line 56: why this one is returning string while the others `operation` ?

packages/massa-web3/src/experimental/basicElements/operationManager.ts
```diff
@@ -0,0 +1,286 @@
...
+
+export interface CallOperation extends BaseSmartContractOperation {
+  type: OperationType.CallSmartContractFunction
+  address: string
+  functionName: string
+  parameter: Uint8Array
+}
+
+export interface ExecuteOperation extends BaseSmartContractOperation {
+  contractDataBinary: Uint8Array
```
AurelienFT, line 66: type set is missing

packages/massa-web3/src/experimental/basicElements/operationManager.ts
```diff
@@ -0,0 +1,286 @@
+import { unsigned } from 'big-varint'
```
AurelienFT, line 1: I think this should be abstracted to not depend from it directly there (we should be able to use the serialize interface)

packages/massa-web3/src/experimental/basicElements/operationManager.ts
```diff
@@ -0,0 +1,286 @@
...
+    let operationDetails: BaseOperation = {
+      fee: decodeNext(),
+      expirePeriod: decodeNext(),
+      type: decodeNext(),
+    }
+
+    switch (operationDetails.type) {
+      case OperationType.Transaction: {
+        const recipientAddress = Address.fromVersionedBytes(
+          data.slice(offset, offset + 33)
```
AurelienFT, line 166: Magic number not clear to me (and not future proof a version could be in more than 1 byte in the future)

packages/massa-web3/src/experimental/basicElements/operationManager.ts
```diff
@@ -0,0 +1,286 @@
...
+    operation: OperationDetails,
+    key: PublicKey
+  ): Uint8Array {
+    // u64ToBytes is little endian
+    const networkId = new Uint8Array(8)
+    const view = new DataView(networkId.buffer)
+    view.setBigUint64(0, BigInt(chainId), false)
+
+    const data = OperationManager.serialize(operation)
+    const publicKeyBytes = key.versionedBytes()
```
AurelienFT, line 212: if we don't have an other `toBytes()` than `versionedBytes` maybe we should rename it to `toBytes()` and use this only representation as byte everywhere

packages/massa-web3/src/experimental/basicElements/address.ts
gregLibert: The source of the problem is the choice made at the node level to use this encoding:
- The base58 encoded bytes contain the version, but not the type of the address.
- During signature, the type of the address must be used (and I don't see any reason to do so).

So, we ended up with a bicephalous solution matching the bicephalous conception...

packages/massa-web3/src/experimental/accountOperation.ts
gregLibert: I don't understand your remark here. Let discuss that during a call.

packages/massa-web3/src/experimental/basicElements/operationManager.ts
gregLibert: Varint is intrinsically part of the serialization and deserialization logic. If we want to offer abstraction, the part that should be abstracted should be the `Canonicalizer` (canonicalization operation including `canonicalize` and `uncanonicalization`, if that makes sense from a code usability perspective).

packages/massa-web3/src/experimental/basicElements/operationManager.ts
gregLibert: i didn't find any good solution for that. Except maybe add const attribute to the address object. :shrug:

packages/massa-web3/src/experimental/basicElements/operationManager.ts
AurelienFT: Yes I agree no need to change for now

packages/massa-web3/src/experimental/basicElements/operationManager.ts
AurelienFT: My point is that there is no other toBytes in this package

packages/massa-web3/src/experimental/basicElements/operationManager.ts
AurelienFT: We could, as we do in the node, always return the rest of the data as well as the value (or the next offset)

packages/massa-web3/src/experimental/basicElements/address.ts
AurelienFT: yeah it's not that important

packages/massa-web3/src/experimental/accountOperation.ts
AurelienFT: Now, buy rolls and sells rolls return a string and tranfer an operation. We can't unify ?

packages/massa-web3/src/experimental/basicElements/operationManager.ts
gregLibert: Sorry for the misunderstanding. When dealing with `toBytes` functions on cryptographic objects, I'm expecting to have something that represents a *real* key, which is not the case here. Having fallen into this pitfall during some implementations, I went with a clearer terminology to help the next developer who works on this.

Furthermore, the end-user developer should never deal with this function directly.

This is why I've decided to offer a more verbose, but clearer API.

packages/massa-web3/src/experimental/basicElements/operationManager.ts
gregLibert: See #564

packages/massa-web3/src/experimental/basicElements/operationManager.ts
```diff
@@ -149,35 +150,37 @@ export class OperationManager {
...
     switch (operationDetails.type) {
       case OperationType.Transaction: {
+        // decode address type to get encoded size
+        varint.decode(data, offset)
         const recipientAddress = Address.fromBytes(
-          data.slice(offset, offset + 33)
+          data.slice(offset, varint.decode.bytes + BLAKE3_HASH_BYTELENGTH)
         )
-        offset += 33
+        offset += varint.decode.bytes + BLAKE3_HASH_BYTELENGTH
```
gregLibert, line 172: This logic is specific to the Address object and should be encapsulated there. Maybe using a `from transaction(data: Uint8Array, offset: number): Address, number`?

packages/massa-web3/src/experimental/basicElements/address.ts
```diff
@@ -4,11 +4,12 @@ import { Version, Versioner } from '../crypto/interfaces/versioner'
 import VarintVersioner from '../crypto/varintVersioner'
 import { checkPrefix, extractData } from './internal'
 import { PublicKey } from './keys'
+import varint from 'varint'
 
 const ADDRESS_PREFIX = 'A'
 const ADDRESS_USER_PREFIX = 'U'
 const ADDRESS_CONTRACT_PREFIX = 'S'
-
+export const BLAKE3_HASH_BYTELENGTH = 32
```
gregLibert, line 12: This name is not well chosen as it exposes the underlying logic of the Address object's functioning. You should either rename it or change the logic (as proposed below) to encapsulate this information.

packages/massa-web3/src/experimental/account.ts
```diff
@@ -112,15 +114,15 @@ export class Account {
...
-          cipheredData: await passwordSeal
+          Nickname: '',
+          Salt: passwordSeal.salt,
+          Nonce: passwordSeal.nonce,
+          CipheredData: await passwordSeal
             .seal(this.privateKey.toBytes())
             .then((a) => Array.from(a)),
-          publicKey: Array.from(this.publicKey.toBytes()),
+          PublicKey: Array.from(this.publicKey.toBytes()),
         } as AccountV0KeyStore
```
gregLibert: @peterjah I don't really know what is the difference between version 0 and version 1 of the wallet. Do you?

packages/massa-web3/src/experimental/account.ts
```diff
@@ -224,26 +226,33 @@ export class Account {
...
-type AccountKeyStore = AccountV0KeyStore | AccountV1KeyStore
+  /**
+   * Uses the environment variables to create an account.
+   *
+   * @returns An account instance.
+   */
+  static async fromYaml(path: string, password: string): Promise<Account> {
+    // check that file exists
+    if (!existsSync(path)) {
+      throw new Error(`Wallet file "${path}" does not exist.`)
```
gregLibert, line 238: In a similar way to what we are doing with Go, I propose starting exception messages with a lowercase letter, as they should be composed. Except if that's not idiomatic in TS.

packages/massa-web3/src/experimental/basicElements/operationManager.ts
peterjah: i renamed the BLAKE3_HASH_BYTELENGTH to UNDERLYING_HASH_LEN and encapsulated the code to get the address lengh in getByteLength static method of Address

PR #563, opened 2024-04-19
560 xp add sc manager

gregLibert:


packages/massa-web3/src/experimental/smartContract.ts
```diff
@@ -0,0 +1,233 @@
...
+        fee: opts?.fee,
+        periodToLive: opts?.periodToLive,
+        coins: opts?.coins,
+        maxGas: opts?.maxGas,
+        datastore,
+      }
+    )
+
+    const event = opts.waitForSpeculativeExecution
+      ? await operation.getSpeculativeEvents()
```
AurelienFT, line 212: I find the boolean misleading. When I read it I think waiting for speculative or not waiting at all. I would prefer something like an enum with no wait, wait speculative, final. If we just want the last two options we can rename it `waitFinalExecution` if false : speculative, if true final

AurelienFT (commented):
It's hard to review this whole PR given the fact that most of the changes are reflected on #559

packages/massa-web3/src/experimental/smartContract.ts
gregLibert: Is the noWait a real use case?

gregLibert: Let's go with renaming then.

peterjah (approved):

PR #571, opened 2024-04-24
[xp] unify address bytes representation

peterjah:


packages/massa-web3/src/experimental/basicElements/address.ts
```diff
@@ -43,58 +45,102 @@ export class Address {
...
+    return varint.decode(this.bytes)
+  }
+
+  /**
+   * Get the address version.
+   *   *
+   * @returns the address type enum.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  static getVersion(data: string | Uint8Array): Version {
```
gregLibert, line 120: I don't see the point of having this function public as it will add complexity. Except if you have an use case in mind.

packages/massa-web3/src/experimental/basicElements/address.ts
```diff
@@ -43,58 +45,102 @@ export class Address {
...
    * @param publicKey - The public key to derive the address from.
    *
    * @returns A new address object.
    */
-  public static fromPublicKey(publicKey: PublicKey): Address {
-    const address = Address.initFromVersion()
-    address.bytes = publicKey.hasher.hash(publicKey.toBytes())
+  public static fromPublicKey(
+    publicKey: PublicKey,
+    version = DEFAULT_VERSION
```
gregLibert, line 135: Why do we need a version here?

packages/massa-web3/src/experimental/basicElements/address.ts
```diff
@@ -43,58 +45,102 @@ export class Address {
...
+  ): Address {
+    const address = Address.initFromVersion(version)
+    const rawBytes = publicKey.hasher.hash(publicKey.toBytes())
+    address.version = version
```
gregLibert, line 139: useless. This is already done by initFromVersion

packages/massa-web3/src/experimental/basicElements/address.ts
```diff
@@ -18,20 +19,21 @@ const UNDERLYING_HASH_LEN = 32
...
- *   - Add a new default version matching the last version.
+ * - Add a new case in the switch statement with the new algorithms to use.
+ * - Change the DEFAULT_VERSION version matching the last version.
+ * - Change the getVersion method to detect the version from user input.
  * - check the `fromPublicKey` method to potentially adapt how an address is derived from a public key.
  * - Voila! The code will automatically handle the new version.
  */
 export class Address {
-  public bytes: Uint8Array
+  private bytes: Uint8Array
```
gregLibert, line 29: Now that the subject is clarified, you should precise that this is the raw data of the address with the version but without its type.

packages/massa-web3/src/experimental/basicElements/address.ts
```diff
@@ -139,18 +180,22 @@ export class Address {
...
+    )
+    return `${
       this.isEOA ? ADDRESS_USER_PREFIX : ADDRESS_CONTRACT_PREFIX
     }${this.serializer.serialize(versionedBytes)}`
   }
 
   /**
-   * Get byte length of address in binary format .
+   * Get address in binary format from a bytes buffer.
+   * The address should be the first element in the buffer.
```
gregLibert, line 194: Why not passing an offset to remove this constraint?

packages/massa-web3/src/experimental/basicElements/address.ts
```diff
@@ -139,18 +180,22 @@ export class Address {
...
    *
-   * @returns The address length in bytes.
+   * @returns The address in bytes format.
    */
-  static getByteLength(data: Uint8Array): number {
+  static extractFromBuffer(data: Uint8Array): Uint8Array {
```
gregLibert, line 198: Don't you think that you should also returned the number of bytes used from this buffer to let the caller continue his potential deserailization steps.

packages/massa-web3/src/experimental/basicElements/internal.ts
```diff
@@ -17,11 +17,12 @@ export function extractData(
   return extractedData
 }
 
-export function checkPrefix(str: string, ...expected: string[]): void {
+export function checkPrefix(str: string, ...expected: string[]): string {
```
gregLibert, line 20: Badly named. If you return the prefix, it should be something like `mustExtractPrefix`. The "must" is there to explicitly indicate that a failure will throw an exception.

packages/massa-web3/test/experimental/unit/address.spec.ts
```diff
@@ -33,8 +32,9 @@ describe('Address tests', () => {
   })
 
   test('fromString throws error for invalid version', () => {
+    const invalidVersion = -1 as Version
     expect(() =>
-      Address.fromString(account.address.toString(), invalidVersion)
+      Address.fromPublicKey(account.publicKey, invalidVersion)
```
gregLibert, line 37: It's not a fromString anymore

gregLibert (changes requested):

packages/massa-web3/src/experimental/basicElements/address.ts
peterjah: I think its the only case where we cannot deduce it ? and we need it to create address bytes

packages/massa-web3/src/experimental/basicElements/address.ts
peterjah: bytes contains address type and version. i ll add a comment

packages/massa-web3/src/experimental/basicElements/address.ts
peterjah: i think it would make the code more complex.
The lenght is easily found using `res.length`

gregLibert: It may be true with the current implementation, but it's not for others (e.g., the array using Args). I've proposed such an interface because it's classic when extracting data from a buffer. Having consistent function signatures is something important, in my humble opinion.

packages/massa-web3/src/experimental/basicElements/address.ts
```diff
@@ -43,58 +46,101 @@ export class Address {
...
     switch (version) {
       case Version.V0:
-        return new Address(new Base58(), new VarintVersioner(), version, false)
+        return new Address(new Base58(), new VarintVersioner(), version)
       default:
         throw new Error(`unsupported version: ${version}`)
     }
   }
 
+  private getPrefix(str: string): string {
```
gregLibert, line 60: Should this function be changed to static since we don't use 'this'? What is the best practice in TypeScript?

packages/massa-web3/src/experimental/basicElements/address.ts
```diff
@@ -43,58 +46,101 @@ export class Address {
...
     } catch (e) {
       throw new Error(`invalid address string: ${e.message}`)
     }
-
     return address
   }
 
+  /**
+   * Get the address type from bytes.
+   *   *
```
gregLibert, line 105:
```suggestion
   * 
```

packages/massa-web3/src/experimental/basicElements/address.ts
```diff
@@ -43,58 +46,101 @@ export class Address {
...
+  private getType(): number {
+    if (!this.bytes) {
+      throw new Error('address bytes is not initialized')
+    }
+    return varint.decode(this.bytes)
+  }
+
+  /**
+   * Get the address version.
+   *   *
```
gregLibert, line 117:
```suggestion
   *
```

packages/massa-web3/src/experimental/basicElements/address.ts
gregLibert: I disagree. This implementation is only valid for a specific public key and address version.
My suggestion is to:
- remove the version argument,
- check that the publicKey version is V0 - otherwise throw an error,
- initFromVersion using V0

And keep the rest of the code.

packages/massa-web3/src/experimental/basicElements/address.ts
```diff
@@ -107,13 +153,10 @@ export class Address {
...
-    const addressType = varint.decode(bytes)
-    const { version, data } = new VarintVersioner().extract(
-      bytes.slice(varint.decode.bytes)
-    )
+    const version = Address.getVersion(bytes)
     const address = Address.initFromVersion(version)
-    address.bytes = data
-    address.isEOA = addressType === 0
+    address.bytes = bytes
+    address.isEOA = address.getType() === 0
```
gregLibert, line 159: Magic number.

packages/massa-web3/src/experimental/basicElements/internal.ts
```diff
@@ -17,11 +17,12 @@ export function extractData(
   return extractedData
 }
 
-export function checkPrefix(str: string, ...expected: string[]): void {
+export function mustExtractPrefix(str: string, ...expected: string[]): string {
```
gregLibert, line 20: Exported function should have a documentation

gregLibert (changes requested):

packages/massa-web3/src/experimental/basicElements/address.ts
peterjah: i am not aware of any best practice about this. lets go for static then it should be totally equivalent

packages/massa-web3/src/experimental/basicElements/internal.ts
peterjah: i aim to delete this function because its implemented with a shortcut (we arbitrary take the length of the first exected prefix), so if expected prefix length are differents, this will lead to wrong behaviour.
but yes ill add ts doc for now

gregLibert (approved):

PR #586, opened 2024-04-30
Fix linter

gregLibert:


packages/massa-web3/src/experimental/basicElements/keys.ts
```diff
@@ -199,6 +200,7 @@ export class PrivateKey {
 export class PublicKey {
   public bytes: Uint8Array
 
+  // eslint-disable-next-line max-params
```
ASAPSegfault, line 203: I feel like we'll need to add this statement on many functions, especially constructors.
Maybe we can set a higher number for max-params but otherwise I would consider removing this rule due to overhead.

packages/massa-web3/src/experimental/basicElements/operationManager.ts
```diff
@@ -24,7 +31,7 @@ export enum OperationType {
  *
  * @remarks
  * Period to live is the number of periods the operation is valid for.
- * This value must be positive and if it's too big, the node will (silently?) reject the operation.
+ * This value must be positive and if it's too big, the node will (sliently?) reject the operation.
```
ASAPSegfault, line 34:
```suggestion
 * This value must be positive and if it's too big, the node will (silently?) reject the operation.
```

packages/massa-web3/src/experimental/basicElements/operationManager.ts
```diff
@@ -110,17 +117,6 @@ export class OperationManager {
...
-        // @see https://docs.massa.net/docs/learn/operation-format-execution#callsc-operation-payload
-        operation = operation as CallOperation
-        components.push(unsigned.encode(operation.maxGas))
-        components.push(unsigned.encode(operation.coins))
-        components.push(Address.fromString(operation.address).toBytes())
-        components.push(varint.encode(operation.func.length))
-        components.push(Buffer.from(operation.func))
-        components.push(varint.encode(operation.parameter.length))
-        components.push(operation.parameter)
-        break
```
ASAPSegfault, line 123: I assume we don't need this code thanks to Ben's refacto on callSC ?

packages/massa-web3/src/experimental/basicElements/keys.ts
gregLibert: You're right, but it's on purpose.
Here, the constructor is protected, which means that an end-user won't call it directly. This is why I have accepted disabling this rule. If the method were public, I would have reorganized the code.
The underlying idea is to have a cleaner SDK by reducing the number of arguments in the method signatures.

packages/massa-web3/src/experimental/basicElements/operationManager.ts
gregLibert: My bad, I'll fix it.

packages/massa-web3/src/experimental/basicElements/internal.ts
```diff
@@ -1,5 +1,6 @@
 import Serializer from '../crypto/interfaces/serializer'
 import { Versioner, Version } from '../crypto/interfaces/versioner'
+import { FIRST, ONE } from '../utils'
```
Ben-Rey, line 3: If One I would prefer ZERO instead of First. When seing this I thought "First what".

Ben-Rey (approved):

packages/massa-web3/src/experimental/basicElements/internal.ts
gregLibert: `FIRST` is first element of an array or a string.

ASAPSegfault (approved):

PR #588, opened 2024-05-02
Lint args and serializers

ASAPSegfault:


gregLibert (approved):
Let's get this merged, we will improve code readibility while reworking on args and serailizer.

PR #600, opened 2024-05-06
599 replace bigint with u64 when needed2

gregLibert:


packages/massa-web3/src/experimental/basicElements/serializers/arrays.ts
```diff
@@ -159,10 +149,10 @@ export function arrayToBytes(
         args.addI64(value as bigint)
         break
       case ArrayTypes.U32:
-        args.addU32(value as number)
+        args.addU32(value as U32.U32)
```
Ben-Rey, line 152: should u32 be a number and not a bigInt?

packages/massa-web3/src/experimental/basicElements/mas.ts
```diff
@@ -19,12 +19,8 @@ export type Mas = bigint
...
  * @throws An error if the value is not a safe integer.
  */
-export function fromMas(value: number): Mas {
-  if (!Number.isSafeInteger(value)) {
-    throw new Error(ERROR_NOT_SAFE_INTEGER)
-  }
-
-  return BigInt(value * POWER_TEN ** NB_DECIMALS)
+export function fromMas(value: U64): Mas {
+  return fromNumber(value * BigInt(POWER_TEN) ** BigInt(NB_DECIMALS))
```
Ben-Rey, line 23: Naming the variable TEN should be enough so we can move it to the no magic number file

packages/massa-web3/src/experimental/basicElements/mas.ts
```diff
@@ -19,12 +19,8 @@ export type Mas = bigint
  *
  * @throws An error if the value is not a safe integer.
  */
-export function fromMas(value: number): Mas {
-  if (!Number.isSafeInteger(value)) {
-    throw new Error(ERROR_NOT_SAFE_INTEGER)
-  }
-
-  return BigInt(value * POWER_TEN ** NB_DECIMALS)
+export function fromMas(value: U64): Mas {
```
Ben-Rey, line 22: Here we return Mas type. Maybe it would be more clear if we return NanoMas

Ben-Rey (approved):

packages/massa-web3/src/experimental/basicElements/serializers/arrays.ts
gregLibert: A number in javascript can be a float. 1.1 is a number. U32 is an integer.

packages/massa-web3/src/experimental/basicElements/mas.ts
gregLibert: Even if TEN and POWER_TEN have the same underlying value, having TEN doesn't explain the reason behind our actions. 
Here, we want to convert the integer value to the expected numerical class in base 10. That’s why we use a power of ten for this conversion.
Similarly, we use variables like milli, micro, and nano to make the code more readable. This approach clarifies the purpose and improves the understanding of the transformations being applied.

packages/massa-web3/src/experimental/basicElements/mas.ts
gregLibert: I don't understand. The main type is the Mas, and currencies, to avoid precision issues, are mostly represented in their minor units. 

That being said, do you want me to add a comment on the Mas type to explain that it is represented in its minor unit, the nano massa?

Ben-Rey: It might be useful to avoid misunderstanding

PR #602, opened 2024-05-09
Read only call

Ben-Rey:


packages/massa-web3/src/experimental/client/interfaces.ts
```diff
@@ -33,6 +61,7 @@ export interface BlockchainClient {
   getEvents(filter: EventFilter): Promise<SCOutputEvent[]>
   getChainId(): Promise<bigint>
   getMinimalFee(): Promise<bigint>
+  executeReadOnlyCall(opt: ReadOnlyCallParams): Promise<ReadOnlyCallResult>
```
gregLibert, line 64: The parameter name is misleading; "opt" stands for optional. Here, nothing is optional.

packages/massa-web3/src/experimental/client/interfaces.ts
```diff
@@ -21,6 +22,33 @@ export type EventFilter = {
...
 }
 
+export type ReadOnlyCallParams = {
+  func: string
+  targetAddress: string
+  callerAddress: string
+  parameter?: Uint8Array
+  coins?: Mas
+  fee?: Mas
+  maxGas?: bigint
```
gregLibert, line 32: Should be a U64

packages/massa-web3/src/experimental/client/publicAPI.ts
```diff
@@ -78,11 +85,48 @@ export class PublicAPI {
...
+      {
+        max_gas: Number(readOnlyCall.maxGas),
+        target_address: readOnlyCall.targetAddress,
+        target_function: readOnlyCall.func,
+        parameter: readOnlyCall.parameter
+          ? Array.from(readOnlyCall.parameter)
+          : [],
+        caller_address: readOnlyCall.callerAddress,
+        coins: readOnlyCall.coins
+          ? Mas.toString(readOnlyCall.coins).padStart(Mas.NB_DECIMALS, '0')
```
gregLibert, line 100: Separation of concerns: it's not the job of the caller to handle padding, it should be done at the Mas level if needed.

packages/massa-web3/src/experimental/client/publicAPI.ts
```diff
@@ -78,11 +85,48 @@ export class PublicAPI {
...
 
   async executeReadOnlyCall(
-    readOnlyCall: ReadOnlyCall
-  ): Promise<ExecuteReadOnlyResponse> {
-    return this.connector
-      .execute_read_only_call([readOnlyCall])
-      .then((r) => r[FIRST])
+    readOnlyCall: ReadOnlyCallParams
+  ): Promise<ReadOnlyCallResult> {
+    const [readOnlyRes] = await this.connector.execute_read_only_call([
```
gregLibert, line 90: Separation of concerns: the level of abstraction chosen for your variable name is incorrect. Using `readOnlyRes` in a *readOnlyCall* is overcomplicated. `res` or `response` should be sufficient.

packages/massa-web3/src/experimental/smartContract.ts
```diff
@@ -7,38 +7,48 @@ import {
...
 export const MAX_GAS_CALL = 4294167295n
 export const MIN_GAS_CALL = 2100000n
+export const MAX_GAS_DEPLOYMENT = 3980167295n
+
 const DEFAULT_PERIODS_TO_LIVE = 9
 
 type CommonOptions = {
-  fee?: bigint
+  fee?: Mas
   maxGas?: bigint
```
gregLibert, line 27: Should be a U64

packages/massa-web3/src/experimental/smartContract.ts
```diff
@@ -7,38 +7,48 @@ import {
...
-  maxCoins?: bigint
+  maxCoins?: Mas
   periodToLive?: number
   waitFinalExecution?: boolean
 }
 
+type ReadOnlyCallOptions = {
+  parameter?: Uint8Array
```
gregLibert, line 41: It should be an argument of the `read` function. Even if it won't always be necessary, you can set a default value if you want, since in most cases we call a function with parameters.

packages/massa-web3/src/experimental/smartContract.ts
```diff
@@ -7,38 +7,48 @@ import {
...
+  callerAddress?: string
+  coins?: Mas
+  fee?: Mas
+  maxGas?: bigint
```
gregLibert, line 45: U64

packages/massa-web3/src/experimental/smartContract.ts
```diff
@@ -192,6 +207,23 @@ export class SmartContract {
...
+    func: string,
+    opts: ReadOnlyCallOptions
+  ): Promise<ReadOnlyCallResult> {
+    opts.maxGas = opts.maxGas ?? MAX_GAS_CALL
+
+    return await this.client.executeReadOnlyCall({
+      func,
+      parameter: opts.parameter,
+      targetAddress: this.contractAddress,
+      callerAddress: opts.callerAddress || this.contractAddress,
```
gregLibert, line 220: I think that the default caller  address should be the wallet owner instead.

packages/massa-web3/src/experimental/smartContract.ts
```diff
@@ -207,13 +239,29 @@ export class SmartContract {
...
+   *
+   * @param func - The function to estimate the gas cost.
+   * @returns The gas estimation for the function.
    */
-  static async getGasEstimation(): Promise<bigint> {
-    return MAX_GAS_CALL
+  async getGasEstimation(
+    func: string,
+    parameter: Uint8Array,
+    callerAddress: Address,
```
gregLibert, line 250: Should be an option IMHO

packages/massa-web3/src/experimental/smartContract.ts
```diff
@@ -207,13 +239,29 @@ export class SmartContract {
...
+    opts: CallOptions
+  ): Promise<bigint> {
```
gregLibert, line 252: Should be a U64

packages/massa-web3/src/experimental/smartContract.ts
```diff
@@ -207,13 +239,29 @@ export class SmartContract {
...
   /**
-   * Estimates the gas required for an operation.
-   * Currently, it returns a predefined maximum gas value.
-   * @returns A promise that resolves to the estimated gas amount in bigint.
-   * TODO: Implement dynamic gas estimation using dry run call.
-   * TODO: Remove static if needed.
+   * Returns the gas estimation for a given function.
+   *
+   * @param func - The function to estimate the gas cost.
+   * @returns The gas estimation for the function.
```
gregLibert, line 245: Throws documentation is missing.

packages/massa-web3/src/experimental/smartContract.ts
```diff
@@ -192,6 +207,23 @@ export class SmartContract {
     return new Operation(this.client, await operation.send(details))
   }
 
+  async read(
```
gregLibert, line 210: Documentation is missing

packages/massa-web3/test/experimental/integration/publicAPI.spec.ts
```diff
@@ -213,17 +214,16 @@ describe('client tests', () => {
     expect(bytecodes).toHaveLength(2)
   })
 
-  test('executeReadOnlyCall', async () => {
+  xtest('executeReadOnlyCall', async () => {
```
gregLibert, line 217: Why is it deactivated?

packages/massa-web3/test/experimental/integration/smartContract.spec.ts
```diff
@@ -167,5 +168,70 @@ describe('Smart Contract', () => {
...
+        const result = await contractTest.read('sendCoins', {
+          coins: coinAmount,
+        })
+
+        expect(result.info.events[0].data).toBe(
+          `Received ${coinAmount.toString()} coins`
+        )
+      })
+
+      xtest('Read only call with maxGas', async () => {
```
gregLibert, line 226: Why is it deactivated?

gregLibert (changes requested):

packages/massa-web3/src/experimental/smartContract.ts
Ben-Rey: Ok in that case we must pass an account or make the callerAddress not optional.

packages/massa-web3/src/experimental/smartContract.ts
Ben-Rey: If no callerAddress the read operation fail

packages/massa-web3/src/experimental/smartContract.ts
gregLibert: But it is for the read function, right?

gregLibert: You're right, this is not a clean alternative. How is it supposed to be used?

packages/massa-web3/src/experimental/smartContract.ts
gregLibert: But the read function has it as an optional parameter.

packages/massa-web3/src/experimental/smartContract.ts
```diff
@@ -207,13 +239,29 @@ export class SmartContract {
...
+      fee: opts.fee,
+      callerAddress: callerAddress.toString(),
+      coins: opts.coins,
+    })
+
+    if (result.info.error) {
+      throw new Error(result.info.error)
+    }
+
+    return BigInt(result.info.gasCost)
```
gregLibert, line 265: Aren't we supposed to have a security marging?

Ben-Rey: Not sure as it was not the case before but I'll ask

packages/massa-web3/src/experimental/basicElements/mas.ts
```diff
@@ -100,25 +100,33 @@ export function fromString(value: string): Mas {
...
-  const integerPart = valueString.slice(FIRST, -NB_DECIMALS) || '0'
-
+  const integerPart = valueString.slice(ZERO, -NB_DECIMALS) || '0'
   let decimalPart = valueString.slice(-NB_DECIMALS).padStart(NB_DECIMALS, '0')
 
-  if (!keepTrailingZeros) {
+  if (decimalPlaces === null) {
     decimalPart = decimalPart.replace(/0+$/, '')
+  } else {
+    // TODO - Round the number to the desired number of decimal places
```
gregLibert, line 127: Changing from truncate to round approximation is indeed a breaking change. Here's how you could handle it:

- If you plan to offer both options, you should add a proper argument to choose between rounding and truncating. If you don't want to spend time on the round implementation, simply throw a `not implemented` exception when it's used.
- If you decide to only offer round approximation, you should clearly document and implement it immediately.
- If you choose to only offer truncate approximation, you should document this  and remove the related TODO comment.

packages/massa-web3/src/experimental/client/publicAPI.ts
```diff
@@ -93,8 +93,8 @@ export class PublicAPI {
         target_function: params.func,
         parameter: params.parameter ? Array.from(params.parameter) : [],
         caller_address: params.callerAddress,
-        coins: params.coins ? Mas.toString(params.coins, true) : null,
-        fee: params.fee ? Mas.toString(params.fee, true) : null,
+        coins: params.coins ? Mas.toString(params.coins) : null,
+        fee: params.fee ? Mas.toString(params.fee) : null,
```
gregLibert, line 97: You've implemented something that is finally not used?

packages/massa-web3/src/experimental/smartContract.ts
```diff
@@ -143,7 +147,7 @@ export class SmartContract {
     const addr = firstEvent.data.split(': ')[ONE]
 
     // TODO: What if multiple smart contracts are deployed in the same operation?
-    return new SmartContract(client, addr)
+    return new SmartContract(client, addr, account)
```
gregLibert, line 150: You should use fromAddress "constructor" instead.

packages/massa-web3/src/experimental/smartContract.ts
```diff
@@ -225,7 +230,7 @@ export class SmartContract {
       func,
       parameter,
       targetAddress: this.contractAddress,
-      callerAddress: opts.callerAddress || this.contractAddress,
+      callerAddress: opts.callerAddress ?? this.account.address.toString(),
```
gregLibert, line 233: Why callerAddress is not an address type? Why setting the type in the name of that field?

packages/massa-web3/test/experimental/integration/setup.ts
```diff
@@ -2,7 +2,7 @@ import 'dotenv/config'
 import { JsonRPCClient } from '../../../src/experimental/client'
 import { Account } from '../../../src/experimental/account'
 import { AccountOperation } from '../../../src/experimental/accountOperation'
-import { toMas } from '../../../src/experimental/utils'
+import * as Mas from '../../../src/experimental/basicElements/mas'
```
gregLibert, line 5: You should import Mas from experimental instead

gregLibert (changes requested):

packages/massa-web3/src/experimental/smartContract.ts
Ben-Rey: you're right

packages/massa-web3/src/experimental/client/publicAPI.ts
Ben-Rey: This differs from what I've been told, but it seems to work like this.

packages/massa-web3/src/experimental/basicElements/mas.ts
```diff
@@ -122,10 +122,23 @@ export function toString(
...
+      // If the rounding digit is 5 or greater, round up
+      decimalPart = (
+        BigInt(decimalPart.slice(ZERO, decimalPlaces)) + BigInt(ONE)
+      )
+        .toString()
+        .padStart(decimalPlaces, '0')
+    } else {
+      // Otherwise, truncate the decimal part to the specified number of places
+      decimalPart = decimalPart.slice(ZERO, decimalPlaces)
+    }
```
gregLibert, line 141: This code is getting to complicated, you should use something like [decimal.js](https://github.com/MikeMcl/decimal.js).

gregLibert (changes requested):

gregLibert (approved):

PR #604, opened 2024-05-16
Fix gas estimation and fix option deploy argument

Ben-Rey:
even when I put default value like waitFinalExecution it does not show on function description when used by dev. So I put default value in ts Doc

### DIFF
--- a/packages/massa-web3/src/experimental/smartContract.ts
+++ b/packages/massa-web3/src/experimental/smartContract.ts
@@ -87,7 +87,13 @@ export class SmartContract {
    * @param client - The client to connect to the desired blockchain.
    * @param account - The account that will deploy the smart contract.
    * @param contract - The contract to deploy.
-   * @param opts - Optional deployment details.
+   * @param opts - Optional deployment details. Default to
+   *  - fee: auto-estimated fee,
+   *  - maxCoins: auto-estimated cost,
+   *  - maxGas: auto-estimated gas,
+   *  - periodToLive: 10,
+   *  - waitFinalExecution: true.
+   *
    *
    * @returns The deployed smart contract.
    *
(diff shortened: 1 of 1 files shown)


packages/massa-web3/src/experimental/smartContract.ts
```diff
@@ -87,7 +87,13 @@ export class SmartContract {
...
    * @param account - The account that will deploy the smart contract.
    * @param contract - The contract to deploy.
-   * @param opts - Optional deployment details.
+   * @param opts - Optional deployment details. Default to
+   *  - fee: auto-estimated fee,
+   *  - maxCoins: auto-estimated cost,
+   *  - maxGas: auto-estimated gas,
+   *  - periodToLive: 10,
+   *  - waitFinalExecution: true.
+   *
```
gregLibert, line 96:
```suggestion
   * @param opts - Optional deployment details with defaults as follows:
   * @param opts.fee - Execution fee, auto-estimated if absent.
   * @param opts.maxCoins - Maximum number of coins to use, auto-estimated if absent.
   * @param opts.maxGas - Maximum executino gas, auto-estimated if absent.
   * @param opts.periodToLive - Duration in blocks before the transaction expires, defaults to 10.
   * @param opts.waitFinalExecution - Whether to wait for the transaction to be finalized, defaults to true.
   *
```

packages/massa-web3/src/experimental/smartContract.ts
```diff
@@ -278,6 +288,7 @@ export class SmartContract {
       throw new Error(result.info.error)
     }
 
-    return BigInt(result.info.gasCost)
+    // eslint-disable-next-line @typescript-eslint/no-magic-numbers
+    return (BigInt(result.info.gasCost) * 120n) / 100n
```
gregLibert, line 292: is result.info.gasCost already a number?
if yes:
```suggestion
    return BigInt(result.info.gasCost * 1.2)
```

gregLibert (approved):

PR #608, opened 2024-05-21
Fix(packages/massa-web3): removed axios to use fetch in BaseClient

0xmemorygrinder:
Hello,

This PR replaces Axios by the native fetch function in order to use this package in the Metamask snap bounty (see Secure ECMAScript requirements).

Thanks in advance for your review.

### DIFF
--- a/packages/massa-web3/src/web3/BaseClient.ts
+++ b/packages/massa-web3/src/web3/BaseClient.ts
@@ -181,40 +181,41 @@ export class BaseClient {
   ): Promise<JsonRpcResponseData<T>> {
     let resp: Response = null
 
-    const body = {
+    const bodyData = {
       jsonrpc: '2.0',
       method: resource,
       params: params,
       id: 0,
     }
-
+    let body;
+    try {
+      body = JSON.stringify(bodyData)
+    } catch (ex) {
+      return {
+        isError: true,
+        result: null,
+        error: new Error('JSON.stringify error: ' + String(ex)),
+      } as JsonRpcResponseData<T>
+    }
     try {
       resp = await fetch(this.getProviderForRpcMethod(resource).url, {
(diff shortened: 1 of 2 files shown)


Ben-Rey (approved):

packages/massa-web3/src/web3/BaseClient.ts
```diff
@@ -192,34 +189,34 @@ export class BaseClient {
...
+      }
+
       return {
-        isError: true,
-        result: null,
-        error: new Error('JSON.parse error: ' + String(ex)),
+        isError: false,
+        result: responseData.result as T,
+        error: null,
       } as JsonRpcResponseData<T>
```
gregLibert, line 212: You can simplify your code by using easy to read ternary operators:
```suggestion
return {
  isError: !!responseData.error,
  result: responseData.error ? null : responseData.result as T,
  error: responseData.error ? new Error(responseData.error.message) : null,
} as JsonRpcResponseData<T>;
```

packages/massa-web3/src/web3/BaseClient.ts
```diff
@@ -192,34 +189,34 @@ export class BaseClient {
...
-
-    const responseData: JsonRpcResponseData<T> = resp.data
-
-    if (responseData.error) {
+    } catch (ex) {
       return {
         isError: true,
         result: null,
-        error: new Error(responseData.error.message),
+        error: new Error('JSON.parse error: ' + String(ex)),
```
gregLibert, line 217: Are you sure that fetch calls never panic?
If not, I suggest to extend the error prefix to something like `'Fetch/JSON error: '`

gregLibert (changes requested):
Nice cleaning!

gregLibert (approved):

PR #623, opened 2024-05-30
Add openrpc generation tests in ci

thomas-senechal:


modship (approved):

open_rpc/src/publicAPI.ts
```diff
@@ -0,0 +1,207 @@
+import {
```
Ben-Rey, line 1: This file already exist in massa-web3 so we should not duplicate it

open_rpc/tests/publicAPI.spec.ts
```diff
@@ -0,0 +1,253 @@
+import {
+  AddressFilter,
```
Ben-Rey, line 2: there is also publicAPI.spec.ts in test/integration/publicAPI.spec.ts. Not sure it's exactly the same

open_rpc/src/publicAPI.ts
Ben-Rey: On "next" branch

PR #640, opened 2024-07-15
update import fs and yaml on browser

Ben-Rey:


src/account.ts
```diff
@@ -2,8 +2,19 @@ import { Version } from './crypto/interfaces/versioner'
...
+
+if (typeof window === 'undefined') {
+  // Running in Node.js
+  import('fs').then((module) => {
+    fs = module
+  })
+  import('js-yaml').then((module) => {
+    yaml = module
+  })
+}
```
peterjah, line 17: please prefer the use of await... 
`fs = await import('fs')`

peterjah (approved):

PR #671, opened 2024-08-23
Add get status to provider

Ben-Rey:


src/utils/formatObjects/nodeStatusInfo.ts
peterjah, line 1: I would put this in a subFolder on the client
`src/client/utils/formatObject.ts`

src/provider/interface.ts
```diff
@@ -38,4 +39,5 @@ export type Provider = {
   deploySC(params: DeploySCParams): Promise<SmartContract>
   getOperationStatus(opId: string): Promise<OperationStatus>
   getEvents(filter: EventFilter): Promise<SCEvent[]>
+  getStatus(): Promise<NodeStatusInfo>
```
peterjah, line 42: maybe "getStatus" is not very well chosen here... maybe getNodeStatus?

test/integration/MRC20.spec.ts
```diff
@@ -59,7 +59,10 @@ describe('Token wrapper tests', () => {
     const amount = 123_000_000n
     let operation = await usdcContract.increaseAllowance(
       'AU1wN8rn4SkwYSTDF3dHFY4U28KtsqKL1NnEjDZhHnHEy6cEQm53',
-      amount
+      amount,
+      {
+        coins: Mas.fromString('0.1'),
+      }
```
peterjah, line 65: i think u needed it fot the first execution with your wallet( but you could probably delete it now.) it will avoid to spend too much coins in the CI wallet

PR #722, opened 2024-12-30
Feat/add public provider

Ben-Rey:


src/basicElements/args.ts
```diff
@@ -133,7 +133,7 @@ export class Args {
    * @param offset - The optional offset to start deserializing from.
    */
   constructor(
-    public serialized = new Uint8Array(),
+    public serialized: Uint8Array = new Uint8Array(),
```
Ben-Rey, line 136: fix: Type 'Uint8Array<ArrayBufferLike>' is not assignable to type 'Uint8Array<ArrayBuffer>
Might not be the best way to fix it

src/smartContracts/smartContract.ts
```diff
@@ -37,7 +38,11 @@ export class SmartContract {
       ...options,
     }
 
-    return this.provider.callSC(callParams)
+    if ('callSC' in this.provider) {
+      return (this.provider as Provider).callSC(callParams)
+    }
+
+    return Promise.reject(new Error('Provider does not support callSC'))
```
peterjah, line 45: why not just `throw new Error('Provider does not support callSC')`?

peterjah: you could have used a typeguard function here:
```
function isProvider(provider: unknown): provider is Provider {
  return 'callSC' in provider;
}
```

src/smartContracts/smartContract.ts
```diff
@@ -37,7 +38,11 @@ export class SmartContract {
       ...options,
     }
 
-    return this.provider.callSC(callParams)
+    if ('callSC' in this.provider) {
```
peterjah, line 41: you could have used a typeguard function here:
```
function isProvider(provider: unknown): provider is Provider {
  return 'callSC' in provider;
}
```

src/provider/web3Provider/web3Provider.ts
```diff
@@ -3,49 +3,78 @@ import {
...
 
-  static mainnet(account: Account): Web3Provider {
-    return Web3Provider.fromRPCUrl(PublicApiUrl.Mainnet, account)
+  static fromRPCUrl(url: string, account: Account): Web3Provider
+  static fromRPCUrl(url: string): PublicProvider
+  static fromRPCUrl(
+    url: string,
+    account?: Account
+  ): Web3Provider | PublicProvider {
+    const client = new PublicAPI(url) // Assuming PublicAPI can be instantiated with a URL
```
peterjah, line 57: `// Assuming PublicAPI can be instantiated with a URL`
rest of chatGPT assistant?

PR #734, opened 2025-01-20
Deferred calls

modship:


peterjah (approved):
gud!
A real e2e test would have been nice to have.

PR #736, opened 2025-02-07
if non existing datastore entry return null

fleandrei:


package.json
```diff
@@ -29,7 +29,7 @@
         "test:browser": "jest --config=jest.browser.config.js",
         "test:cov": "jest --detectOpenHandles --coverage --forceExit",
         "test:cov:watch": "jest --detectOpenHandles --coverage --watch",
-        "test:integration": "jest --forceExit --runInBand ./test/integration",
+        "test:integration": "jest --detectOpenHandles --forceExit --runInBand ./test/integration",
```
peterjah, line 32: i think you can remove detectOpenHandles

peterjah (approved):

PR #742, opened 2025-03-03
GRPC provider - Main.2.5

modship:


modship:
@Ben-Rey @peterjah  for the review you can skip `generated` directory

modship:
I updated the version to 5.2.0 what do you think? @peterjah @Ben-Rey

scripts/proto_build.sh
```diff
@@ -0,0 +1,66 @@
...
+
+echo "Running proto_build script..."
+
+#Variables
+REPO_URL="https://github.com/massalabs/massa-proto" 
+TARGET_DIR="proto"  
+DESTINATION="tmp" 
+BRANCH="main"  
+
+rm -rf tmp/proto
```
peterjah, line 11: rm -rf ${DESTINATION}/${TARGET_DIR}

src/provider/grpcProvider/grpcPublicProvider.ts
```diff
@@ -0,0 +1,708 @@
...
+    return new GrpcPublicProvider(new PublicServiceClient(url), url)
+  }
+
+  /**
+   * Executes a state query on the blockchain
+   */
+  async queryState(
+    queries: ExecutionQueryRequestItem[]
+  ): Promise<QueryStateResponse> {
+    return await this.client.queryState(
```
peterjah, line 99: await not needed here

src/provider/interface.ts
```diff
@@ -21,8 +23,8 @@ export type PublicProvider = {
   networkInfos(): Promise<Network>
   getOperationStatus(opId: string): Promise<OperationStatus>
   getEvents(filter: EventFilter): Promise<t.OutputEvents>
-  getNodeStatus(): Promise<NodeStatusInfo>
-  readSC(params: ReadSCParams): Promise<ReadSCData>
+  getNodeStatus(): Promise<NodeStatusInfo | PublicStatus>
+  readSC(params: ReadSCParams): Promise<ReadSCData | ReadOnlyExecutionOutput>
```
peterjah, line 27: It will be pretty annoying to have to different return type possible.
It is really needed ?
two solution to this:
- modify `getNodeStatus`and `readSC`so they return `NodeStatusInfo`and `ReadSCData`
- modify `NodeStatusInfo`and `ReadSCData` to be more generic and fit the new needs

src/smartContracts/smartContract.ts
```diff
@@ -64,7 +64,7 @@ export class SmartContract {
       target: this.address,
       ...options,
     }
-    return this.provider.readSC(readParams)
+    return this.provider.readSC(readParams) as Promise<ReadSCData>
```
peterjah, line 67: this cast could be removed if addressing the previous comment

test/integration/provider.spec.ts
```diff
@@ -1,10 +1,26 @@
...
+    expect(status.getCurrentCycle()).toBeDefined()
+    expect(status.getCurrentTime()).toBeDefined()
+    expect(status.getCurrentCycleTime()).toBeDefined()
+    expect(status.getLastExecutedFinalSlot()).toBeDefined()
+    expect(status.getLastExecutedSpeculativeSlot()).toBeDefined()
+    expect(status.getFinalStateFingerprint()).toBeDefined()
+    expect(status.getChainId()).toBeDefined()
+    expect(status.getMinimalFees()).toBeDefined()
+  })
+})
```
peterjah, line 21: It would be nice to have more tests. I suggest to put them in a dedicated file. this can be done in a followup issue

scripts/proto_build.sh
modship: done

src/provider/grpcProvider/grpcPublicProvider.ts
modship: done

package.json
```diff
@@ -19,6 +19,7 @@
         "clean-dist": "rimraf dist/*",
         "build-esm": "tsc --project tsconfig.esm.json",
         "build-commonjs": "tsc --project tsconfig.commonjs.json",
+        "build-copy": "cp -r src/generated/grpc/ dist/esm/generated/grpc/ && cp -r src/generated/grpc/ dist/cmd/generated/grpc/",
```
peterjah, line 22: whay is this? seems not used

src/smartContracts/smartContract.ts
peterjah: so we can remove the cast ? ^^

src/client/types.ts
```diff
@@ -18,6 +19,7 @@ export type EventFilter = {
   operationId?: string
   isFinal?: boolean
   isError?: boolean
+  status?: ScExecutionEventStatus
```
peterjah, line 22: this new "status" property is misleading, because it will only work for grpc..
The best solution would be to set it depending on "isFinal" value in the `getEvents`of grpcPulblicProvider.
about `ScExecutionEventStatus`i dont really get why we have a "Readonly" I don't think its relevant in getEvents endpoint.
so to resume:
```
let status = ScExecutionEventStatus.SC_EXECUTION_EVENT_STATUS_CANDIDATE
if(isFinal)
  status = ScExecutionEventStatus.SC_EXECUTION_EVENT_STATUS_FINAL

File: src/provider/types.ts [hunk 8]
```diff
@@ -120,3 +126,40 @@ export type ExecutionStats = {
...
+
+export type ExecutedOpsChangesFilter = {
+  empty?: boolean
+  operationId?: string
+}
+
+export type LedgerChangesFilter = {
+  empty?: boolean
+  address?: string
+}
```
peterjah, line 165: can you move theses types in a dedicated file? like create a `type.ts` or `filtersTypes.ts` in grpc folder

src/provider/grpcProvider/grpcProvider.ts
```diff
@@ -0,0 +1,281 @@
...
+  ) {
+    super(client, url)
+  }
+
+  static buildnet(account: Account): GrpcProvider
+  static buildnet(): GrpcPublicProvider
+  static buildnet(account?: Account): GrpcProvider | GrpcPublicProvider {
+    return account
+      ? GrpcProvider.fromGrpcUrl(GrpcApiUrl.Buildnet, account)
+      : GrpcProvider.fromGrpcUrl(GrpcApiUrl.Buildnet)
```
peterjah, line 52: just `return GrpcProvider.fromGrpcUrl(GrpcApiUrl.Buildnet, account)`should work

src/provider/grpcProvider/grpcProvider.ts
```diff
@@ -0,0 +1,281 @@
...
+  }
+
```
peterjah, line 54: mainnet initializers are missing ?

peterjah (changes requested):

package.json
modship: files should be in dist and when build without that, they wasn"t include

src/client/types.ts
modship: done

src/provider/types.ts
modship: this file was not in jsonRpc folders

src/provider/grpcProvider/grpcProvider.ts
modship: nop

src/provider/grpcProvider/grpcProvider.ts
modship: yes ^^ good catch !!!

modship: done

src/smartContracts/smartContract.ts
modship: my bad yes

test/integration/provider.spec.ts
modship: done

test/integration/grpcProvider.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+      .networkInfos()
+      .then((info) => {
+        expect(info.name).toBeDefined()
+        expect(info.chainId).toBeDefined()
+        expect(info.url).toBeDefined()
+      })
+      .catch((err) => {
+        // possible error with sandboxed node
+        expect(err).toBeDefined()
+        expect(err.message).toContain('Unknown chain id')
```
peterjah, line 58: what does it means ?? 
Tests must be reproducible.
The other problem here is that you catch the error potentially thrown by the 'expect'. so your test will pass even if the expected condition are not met.
I already fixed the same thing for getNodeStatus

modship: removed the sandbox part

package.json
```diff
@@ -84,6 +86,7 @@
         "eslint-config-prettier": "^9.1.0",
         "eslint-plugin-jsdoc": "^48.0.4",
         "eslint-plugin-prettier": "^5.1.3",
+        "google-protobuf": "^3.21.4",
```
peterjah, line 89: this should be in direct dependency because its used in generated proto code

package.json
```diff
@@ -73,6 +74,7 @@
     "devDependencies": {
         "@babel/preset-env": "^7.22.15",
         "@open-rpc/typings": "^1.12.4",
+        "@protobuf-ts/plugin": "^2.9.4",
```
peterjah, line 77: is it needed ?? it seems not used

test/integration/grpcProvider.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+    const stakers: Array<StakerEntry> = await grpcProvider.getStakers()
+    expect(stakers).toBeDefined()
+    expect(stakers[0].getRolls()).toBeGreaterThan(0)
+  })
+
+  test('networkInfos', async () => {
+    const resp = await grpcProvider.networkInfos()
+    expect(resp.name).toBeDefined()
+    expect(resp.chainId).toBeDefined()
+    expect(resp.url).toBeDefined()
```
peterjah, line 51: can you go a little bit further on testing?
You can retrieve theses values from json rpc and compare

peterjah (approved):

PR #745, opened 2025-03-13
linking to decentralized websites

fleandrei:


src/deweb/mns.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  const mnsDomain = mns.replace('.massa', '')
+
+  /* -- Check if a localhost deweb provider is available -- */
+  try {
+    const response = await fetch('http://localhost:8080/__deweb_info')
+    if (await isDewebInfoData(response)) {
+      return `http://${mnsDomain}.localhost:8080${mnsPath}${mnsSearch}`
+    }
+    console.log('No local deweb provider found')
```
peterjah, line 42: wrong log msg

src/deweb/mns.ts
```diff
@@ -0,0 +1,87 @@
...
+  } catch (error) {
+    console.log('No local deweb provider found')
```
peterjah, line 44: if you really want to log the error, use console.error

src/deweb/mns.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  /* -- Check if the current domain is a Deweb provider -- */
+  // retrieve current url informations
+  const { host: currentHost, protocol: currentProtocol } = window.location
+  const subdomains = currentHost.split('.')
+
+  // If the current domain is not a Deweb provider, we redirect to the default URL
+  const defaultUrl = `${DEWEB_REDIRECT_URL}?domain="${uri.split('://')[1]}"`
```
peterjah, line 53: `domain="${mns}"` ?

src/deweb/mns.ts
```diff
@@ -0,0 +1,87 @@
+const DEWEB_REDIRECT_URL = 'https://deweb.massa.network/deweb_redirect'
+
+export async function resolveDeweb(uri: string): Promise<string> {
+  if (typeof window === 'undefined') {
+    throw new Error('This function can only be used in a browser environment')
+  }
```
peterjah, line 6: why ? is there a technical constraint about it ?

src/deweb/mns.ts
fleandrei: In fact it's not really an error. We just check if there is a local provider. If not it's not an error.

src/deweb/mns.ts
fleandrei: In fact it's ${mns}${mnsPath}${mnsSearch}

src/deweb/mns.ts
fleandrei: This function allows to have in classic and deweb websites, links to deweb website.
window is needed to get the domain of the current website and check if it is a deweb provider.

src/deweb/mns.ts
fleandrei: This msg is logged if the response.ok is false or if the returned data is not well formatted. This means that there is no local deweb provider available.

peterjah (approved):

PR #763, opened 2025-05-05
add immutable and fix mns link resolve

fleandrei:


src/deweb/immutable.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  await (waitFinal
+    ? operation.waitFinalExecution()
+    : operation.waitSpeculativeExecution())
+  return operation
+}
+
+export async function isImmutable(
+  address: string,
+  nodeUrl: string,
```
peterjah, line 23: its strange to provide an url here.
It would best to use the provider in function parameter, and get the url from it ( using networkInfos). you can also handle the case where nodeUrl is not available (its the case for bearby) and use the public one based on the chain ID

src/deweb/immutable.ts
```diff
@@ -0,0 +1,32 @@
...
+  waitFinal = false
```
peterjah, line 24: i would propose rename`waitFinal` => `final`, there is no waiting here

peterjah (approved):

peterjah (approved):
Nicey !!

src/deweb/immutable.ts
```diff
@@ -0,0 +1,45 @@
...
+  let nodeUrl = networkInfo.url
+
+  if (!nodeUrl) {
+    nodeUrl = getPublicApiByChainId(networkInfo.chainId)
+    if (!nodeUrl) {
+      throw new Error(`Unknown network chainId: ${networkInfo.chainId}`)
+    }
+  }
+
+  console.log('nodeUrl', nodeUrl)
```
peterjah, line 37: console.log to delete

peterjah (approved):

PR #774, opened 2025-07-02
Implement getAddressesDatastoreKeys method in PublicAPI and add related tests. Update package.json for integration tests and adjust type definitions for datastore keys.

Ben-Rey:


src/client/utils.ts
```diff
@@ -0,0 +1,8 @@
+import { strToBytes } from '../basicElements'
+
+export function formatDatastoreKey(key: Uint8Array | string): Uint8Array {
+  if (typeof key === 'string') {
+    return strToBytes(key)
+  }
+  return key
+}
```
peterjah, line 8: can you move this in `src/utils/formatter.ts`
It is the same idea of `parseCallArgs` and it could be used in different contexts

PR #791, opened 2025-09-09
add areimmutable and multi address mns retrieve

fleandrei:


src/deweb/immutable.ts
```diff
@@ -19,6 +19,35 @@ export async function makeImmutable(
...
+    return []
+  }
+  const networkInfo = await provider.networkInfos()
+  const url = networkInfo.url
+    ? networkInfo.url
+    : getPublicApiByChainId(networkInfo.chainId)
+  if (!url) {
+    throw new Error('Provider does not have a node URL')
+  }
+  const publicAPI = new PublicAPI(url)
```
peterjah, line 41: use method you created => PublicAPI.fromProvider(

src/contracts-wrappers/mns.ts
```diff
@@ -95,6 +101,43 @@ export class MNS extends SmartContract {
...
+
+    const results = await batchListAndCall(
+      addresses,
+      async (addressesBatch) => {
+        // Use the PublicAPI to get datastore keys for multiple addresses in a single call
+        return publicAPI.executeMultipleReadOnlyCall(
+          addressesBatch.map((address) => ({
+            target: this.address,
+            func: 'dnsReverseResolve',
+            caller: (this.provider as Provider).address,
```
peterjah, line 130: you should be able to remove "as Provider" if using isProvider previously

PR #792, opened 2025-09-10
fix doc due to too long alias type in generated code

fleandrei:


peterjah (changes requested):
use typedoc configuration

